PR #58, opened 2019-09-19
Replace spaces in img url as %20

loicvh:
It seems that the images are not loaded if there is a space in the name. [Example of a recipe with the problem.](https://www.colruyt.be/fr/recette/chipolatas-de-campagne-et-pennes-au-celeri)

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -705,8 +705,8 @@ class RecipeService {
         $recipe_json = $this->parseRecipeFile($this->getRecipeFileById($id));
 
         if(!isset($recipe_json['image']) || !$recipe_json['image']) { throw new \Exception('No image specified in recipe'); }  
-
-        $recipe_image_data = file_get_contents($recipe_json['image']);
+	
+	$recipe_image_data = file_get_contents(str_replace(' ', '%20', $recipe_json['image']));
 
         if(!$recipe_image_data) { throw new \Exception('Could not fetch image from ' . $recipe_json['image']); }
 


brogers-navex:
Not the maintainer, but here are my thoughts.

Perhaps the site fixed the issue? I see no spaces in the image URL for the recipe listed. I do see an apostrophe, which could cause an issue, but no spaces.

I think this code looks pretty safe and there wouldn't be that much in terms of unexpected bugs since it's strictly limited to spaces. I can potentially see an issue if a `+` character is required instead of `%20`, but I don't run into many of those. 

However, I can't help but think this issue falls under a larger scope than just space characters. There are two directions this project can go:

1. We should assume that all Image URLs are already valid and error if there is an issue.
2. We should assume that all Image URLs are invalid, and perform a URLENCODE on the URI of the image location.

In my opinion, if your image isn't loading because the URL isn't using the correct formatting/encoding, then the origin needs to be fixed. Have you considered copying the photo to a local directory and simplifying the name, then updating the recipe?

loicvh:
> Perhaps the site fixed the issue? I see no spaces in the image URL for the recipe listed. I do see an apostrophe, which could cause an issue, but no spaces.

Did you try to import the above recipe via cookbook? In my experience, the image is not loaded and I get the following error:
`"Could not fetch image from https://static.colruyt.be/culinair/Folderrecepten VI 1819 + WM 1919/Boerenchipolata’s en romige penne met selder_MR.jpg"`

Some browsers, e.g, Chromium, automatically fill the space with '%20'. This could explain why you did not see it. If you load it with Firefox, it will fill the space with '%20', load the image, and then replace it with the initial url.

> However, I can't help but think this issue falls under a larger scope than just space characters. There are two directions this project can go:
> 
> 1. We should assume that all Image URLs are already valid and error if there is an issue.
> 2. We should assume that all Image URLs are invalid, and perform a URLENCODE on the URI of tthe image location.

I do agree, and I think 2 should be the way to go. Sending an error is just useless because the user has _a priori_ no way to change the recipe URL. Using spaces in URL is probably a bad practice but some people use it and the program should accommodate the user and not vice-versa.

> In my opinion, if your image isn't loading because the URL isn't using the correct formatting/encoding, then **the origin needs to be fixed**. Have you considered copying the photo to a local directory and simplifying the name, then updating the recipe?

Maybe I don't get what you mean here, do you suggest to send an email to the aforementioned site to ask them to fix the URL? Because this battle is already lost. I could indeed update the picture as you propose but this is not really in the spirit of cookbook. The beauty is to simply copy-paste the URL and have your recipe on hand. If I have to manually edit the recipe to have the picture, this is a bit annoying.

I know that it is _only_ the picture which is not essential but it looks better when you want to choose the recipe.

Anyway, thank you for the feedback! I guess it should be better to correct the URL upstream, and also correct the URL of the recipe itself; if some barbarous people use spaces there.

brogers-navex:
> Maybe I don't get what you mean here, do you suggest to send an email to the aforementioned site to ask them to fix the URL? Because this battle is already lost. I could indeed update the picture as you propose but this is not really in the spirit of cookbook. The beauty is to simply copy-paste the URL and have your recipe on hand. If I have to manually edit the recipe to have the picture, this is a bit annoying.

I suppose I'm looking at it from a different point of view. In my view - Nextcloud Cookbook is foremost a self-hosted recipe archive. By using external links for images, you now have a dependency on the original source.

The alternative would be to scrape the resource and host it locally. However, this _could_ be problematic due to copyright, trademarks, etc. So I understand if such feature isn't ever implemented.

Would love to hear from the maintainer on that, actually. Sounds like a fun project if there aren't such restrictions.

mrzapp:
@brogers-navex The images are already cached locally, so making it an explicit feature could make sense. I'm not against it, but we'd have to restructure the storage a bit, because right now all recipes are just stored as JSON files in a single folder.

brogers-navex:
@mrzapp - I'll have to read the code then. I didn't recognize any sort of caching mechanism on my configuration at a glance.

I have some ideas that wouldn't alter the structure of the storage. 

Don't want to post at length here, though. I'll write up an issue on my personal account and go from there.

mrzapp:
@brogers-navex we've already made the structural change, so no need to pursue that :) However, I do like your input about urlencode. If you'd like to contribute a fix, the relevant code is [here](https://github.com/mrzapp/nextcloud-cookbook/blob/develop/lib/Service/RecipeService.php#L754)

PR #66, opened 2019-10-07
Basic REST API implementation

danielroehrig:
Index/List, add, delete, update should work. Also getting the recipe image. I tried to stay as close to the [tutorial](https://docs.nextcloud.com/server/16/developer_manual/app/tutorial.html) as I could. Some routes ... I can't make sense of them (test, find, keywords) and would remove them. I couldn't get the tests running, although I could run tests of other apps. I think you are doing something there, that might only work on your machine at the moment. ;)


mrzapp:
Thanks for your politeness about my implementation, but I'll come clean here: I'm a seasoned developer, and I usually have an obsession with doing things in a standards-compliant way, but I'll have to admit that my implementation here was really lazy. There are no tests, and the server-side rendering was a leftover from the prototype.

If I knew how much traction this app would suddenly get, there is no way I would have gone about it in such a half-assed manner. I apologise for the inconvenience, as I can see you've put a lot of effort into this pull request.

I've merged the changes into master now. Again, thank you for your contribution, it's very valuable!

mrzapp:
About the /test, /find and /keywords routes, you are right, they are not in use anymore. I'll remove them.

danielroehrig:
You created something, that other people enjoy. Gaining unexpected traction is kind of a great problem to have. If you don't mind, I would continue working on on this app. Is there a development branch that could be forked? It feels weird to work on the master branch.

mrzapp:
@danielroehrig Sure, I'll make you a contributor and divide branches into "develop" and "stable"

Issue #75, opened 2019-10-10
references to nonexistent master branch


geeseven:
Greetings,

Currently, there are some dead links in the main and translationfiles README.md files that refer to a nonexistent master branch.  A search shows there are [more references to a master branch][0].  I was about to submit a pull request with a simple `s/master/develop`, but unsure which branch is needed for the automation scripts.  

[0]: https://github.com/mrzapp/nextcloud-cookbook/search?q=master&type=Code

christianlupus:
This is due to the fact that recently the master branch was renamed to develop and a new stable branch was created to simplify collaboration.

A plain regex based replacement will not suffice. E.g. the file `.travis.yml` links to the PostgreSQL version to be used for testing. Although testing is not implemented yet, I did not check every other occurence of the term `master`.

geeseven:
Thanks for the timely update and confirming my concerns about a blanket change. You were also quick to submit a pull request, nice.

Issue #73, opened 2019-10-09
Automatic testing implementation


mrzapp:
As discussed in #71

mrzapp:
@danielroehrig should this task be considered done at this point?

PR #76, opened 2019-10-11
Removing any reference to the master branch in documentations' links

christianlupus:
Triggered by #75.

@mrzapp There is a change in `appinfo/info.xml` with a broken link to an old image URL. I do not know if this will break the presentation in the appstore....

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ They probably haven't been indexed. Try clicking the Settings > Rescan library b
 A lot of websites are unfortunately not following the schema.org/Recipe standard, which makes their recipes impossible to read by this app.
 
 #### A website using correct schema.org markup is not being read correctly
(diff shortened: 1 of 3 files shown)


mrzapp:
It looks good to me, thanks for the pull request. BTW, since you're a collaborator now, feel free to commit changes directly to the `develop` branch. Every pull request you've submitted has been solid, so I don't think my review is needed.

christianlupus:
> It looks good to me, thanks for the pull request. BTW, since you're a collaborator now, feel free to commit changes directly to the `develop` branch. Every pull request you've submitted has been solid, so I don't think my review is needed.

I would have done so. I just wanted to inform you, that the screenshot is not visible on https://apps.nextcloud.com/apps/cookbook at the moment as the link points to the master branch thus results in a 404 error.

After #73 is in place, the self-merging onto `develop` will be more secure. :-)

PR #78, opened 2019-10-15
German translation of "All"

theonlytruth:
The word "All" in the list of keywords / categories wasn't translated. I hope, this is the right way doing it.

### DIFF
--- a/translationfiles/de/cookbook.po
+++ b/translationfiles/de/cookbook.po
@@ -188,6 +188,10 @@ msgstr "Rezept herunterladen"
 msgid "Search"
 msgstr "Suche"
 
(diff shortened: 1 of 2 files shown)


christianlupus:
Hello @theonlytruth. Thank you for submitting the patch. I think this is very well ok.

Just out of curiosity: Did you just add the changed manually in the po files or did you run the `update-pot.sh` file?

theonlytruth:
I only changed it manually. Now - a little late - I red the descriptions about translations. So someone should update the cookbook.pot using the update-pot.sh tool. At the moment the template and other languages doesn't have this string in it.

christianlupus:
I did so a 2 hours ago. There were still some texts untranslated even in de and de_DE. I added these. The other languages need to be updated sooner or later.

PR #84, opened 2019-10-17
Update keyword handling.

dmbreakey:
My editor ("Atom") apparently automatically cleaned up extraneous whitespace. My edits were strictly limited to lines 138-150.

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -104,7 +104,7 @@ class RecipeService {
                         $this_image_size = 0;
 
                         foreach($image_matches as $image_match) {
-                            $this_image_size += (int) $image_match; 
+                            $this_image_size += (int) $image_match;
                         }
 
                         if($image_size === 0 || $this_image_size > $image_size) {
@@ -122,7 +122,7 @@ class RecipeService {
         // Clean up the image URL string
         $json['image'] = stripslashes($json['image']);
 
-        // Make sure that "recipeYield" is an integer which is at least 1 
+        // Make sure that "recipeYield" is an integer which is at least 1
         if(isset($json['recipeYield']) && $json['recipeYield']) {
             $yield = filter_var($json['recipeYield'], FILTER_SANITIZE_NUMBER_INT);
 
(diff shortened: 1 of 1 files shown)


lib/Service/RecipeService.php
```diff
@@ -137,13 +137,17 @@ public function checkRecipe($json) {
...
-            $keywords = $json['keywords'];
-            $keywords = strip_tags($keywords);
-            $keywords = str_replace(',,', ',', $keywords);
-            $keywords = explode(',', $keywords);
-            $keywords = array_unique($keywords);
-            $keywords = implode(',', $keywords);
-            $json['keywords'] = $keywords;
+          $keywords = trim($json['keywords']);
+          $keywords = trim($keywords, ',');
+          $keywords = trim($keywords)
```
christianlupus: This could have been simplified into one line by `trim($json['keywords'], " \t\n\r\o\x0B,");`. But it does not hurt like this.

christianlupus:
I personally would have preferred if you restriced the changes in the PR only to the relevant lines. This makes both the review process faster (I have to check that you did not by accident change anything vital). But also, your editor removed the trailing whitespaces on empty lines which is in fact their identation. I had to work with such code in the past and it is nasty as you have to correct the identation manually at quite a few places. I will try to revert these parts from history, please wait a few minutes.

christianlupus:
I am unable to push the modifications to the PR. Did you [allow modifications by the maintainers](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-forkhttps://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

christianlupus:
I was able to merge using command lineas no other modifications were needed.

lib/Service/RecipeService.php
dmbreakey: I actually tried that, but I think I see what I did wrong. I think I used single-quotes instead of double. Sigh.

dmbreakey:
In regards to whitespace trimming: as far as I could tell, it didn't look like my editor had affected anything that should have been significant, but clearly that wasn't the case. I'll review and try to figure out how to turn that behavior off.

lib/Service/RecipeService.php
christianlupus: Yes, then PHP does interpret fewer symbols in a special way. Similar to variable expansion in double quoted strings while not in single quoted strings.

christianlupus:
It is not necessary that you alter your editor's settings. Depending on your way of commiting in git (command line vs. GUI) you can select only a few lines or a single context into the stageing area. Thus, only these changes are inserted into the commit.

Personally I use `git gui` to do it but you can as well use `git add --interactive`.

dmbreakey:
I'm currently using the [Atom](https://atom.io/) editor, with it's built-in Git and GitHub integration. That said I've only been using it for a short time, so it's possible I've got something misconfigured.

For now I've been letting this integration use whatever it's defaults are, but perhaps I should look at them more closely.

I can't say I'm keen on manipulating Git from the CLI, which is very odd; I **live** at the command-line. But Git--for now--is just confusing, and I'm never sure if I'm doing it right.

christianlupus:
> I can't say I'm keen on manipulating Git from the CLI, which is very odd; I **live** at the command-line. But Git--for now--is just confusing, and I'm never sure if I'm doing it right.

With a :smirk: I can point you to [tis page](https://learngitbranching.js.org). It's playful.

PR #86, opened 2019-10-19
new strings translated to French

sephrat:


christianlupus:
@sephrat thank you very much for your help. Merci bien

PR #87, opened 2019-10-19
Refined keyword processing.

dmbreakey:
Had a chance to review my changes a bit more carefully and was able to write it a bit more efficiently; also integrated suggested fix for initial trim(); thank you @christianlupus 

Also, I think I figured out how to restrict my editor to just my direct edits, rather than the entire file; let me know if that doesn't work.


dmbreakey:
Ack. Found a corner-case in handling commas that I'm working on. Can I withdraw the pull?

christianlupus:
OK, first of all, why do you want to withdraw the PR? Because you see no more benefit in it or because you think you made a mistake? You can click on close at the button and the PR will be marked as closed so no longer pop up.

If you simply need time to work on the thing, this is no problem. We will wait until you are finished until we merge into the main development branch.

Two more things I want to give you on your way:
1. It is a good idea to create a local branch in your local fork for any modification you want to do. You might want to read [this introduction on different workflows](https://www.atlassian.com/git/tutorials/comparing-workflows).
2. As a result of this first point, you made a minor mistake: You worked on a old version of the repo. This was due to the fact, that you did not merge back the changes in this (original) repo to your fork. This causes this error message:
Screenshot_20191019_185758 Whenever this message pups up, this indicates a problem with your work. No problem yet, but to get it merged there is still work needed.

If you have any general questions, you can contact me directly (see mail address in git commits from me).

dmbreakey:
OK, I believe I've resolved it. Do you want me to remerge the changes, then repull?

christianlupus:
Just for reference in this PR: We wrote a few mails in order to sort out the author's repo and bring it back in sync. (General git questions)

dmbreakey:
I think I've got it cleaned up and straightened out now. Resynced my fork, following the instructions from [here](https://github.com/KirstieJane/STEMMRoleModels/wiki/Syncing-your-fork-to-the-original-repository-via-the-browser), and then resolved the merge conflict. Looks like the PR is referencing everything correctly, so I'm pretty sure it's good to go.

Thank you for being patient with me; hopefully I can keep from making silly mistakes from here.

mrzapp:
@danielroehrig since you are in the process of breaking up the RecipeService class into smaller ones, it might be a good idea for you to approve this one.

lib/Service/RecipeService.php
```diff
@@ -137,13 +137,10 @@ public function checkRecipe($json) {
 
         // Make sure that "keywords" is an array of unique strings
         if(isset($json['keywords']) && is_string($json['keywords'])) {
-          $keywords = trim($json['keywords']);
-          $keywords = trim($keywords, ',');
-          $keywords = trim($keywords)
+          $keywords = trim($json['keywords'], " \0\t\n\x0B\r,");
```
danielroehrig: I think, this is the same as 
```suggestion
          $keywords = trim($json['keywords']);
```
[trim documentation](https://www.php.net/manual/en/function.trim.php)

dmbreakey: No.

If you notice, we're adding a comma to the list of characters to be trimmed; otherwise yes, it would be the same.

danielroehrig:
Let's hope I pressed the right buttons. ;)

PR #88, opened 2019-10-20
Add dutch translations

Timvde:
Small question: is there really an important difference between "Prep time" and "Preparation time"? Does it need to be shorter? Because I translated them the same... (Idem for "Cook time" vs "Cooking time")

### DIFF
--- a/translationfiles/README.md
+++ b/translationfiles/README.md
@@ -1,6 +1,6 @@
 # Translations for the cookbook app
 
-This is a quick instruction how to generate translations for the nextcoud app `cookbook`.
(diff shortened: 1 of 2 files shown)


christianlupus:
About your question regarding abbreviated versions ("prep time"): I am not perfectly sure where these are popping up (did not check). Could be the case, that the abreviated verion was just inserted as a common term by a different programmer. I do not think this will cause a problem here. Although it might be a good idea to have only one text...

I will merge this PR anyways as it looks good.

Thanks for contributing.

PR #90, opened 2019-10-21
Use dependency injection to get rid of the RecipeService Constructor

danielroehrig:
That makes it easier to test.


danielroehrig:
So, while trying to get testing to work, I stumbled upon a "new" constructor call in the Page- and RecipeController for the RecipeService and it's not necessary, since NextCloud has a DI-container going that can be used instead. That would also make testing easier/possible. Is that okay with you @mrzapp?

mrzapp:
@danielroehrig absolutely, it was just out of ignorance anyway

PR #93, opened 2019-10-23
Removed unused variables

danielroehrig:
I searched the templates for any usage of the "all_recipes" or the
"current_node" variable and did find nothing. Deleting them has no
noticeable effect.

### DIFF
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -37,12 +37,10 @@ class PageController extends Controller
     public function index()
     {
         $view_data = [
-            'all_recipes' => $this->service->getAllRecipesInSearchIndex(),
             'all_keywords' => $this->service->getAllKeywordsInSearchIndex(),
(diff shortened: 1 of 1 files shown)


danielroehrig:
Sometimes, I stumble upon code that I simply don't understand. Maybe if we could have a slack channel or similar, we could talk about them instead of creating a gazillion pull requests. ;)

mrzapp:
@danielroehrig There may still be leftover code examples from the app tutorial, or from previous stuctural design irerations. 

I've created a matrix.org channel here: https://matrix.to/#/#nextcloud-cookbook:matrix.org

danielroehrig:
Cool. My username is danielroehrig and the room seems to be invite only. Can you make the room public or send me an invite?

mrzapp:
@danielroehrig the room is public, but I invited you anyway :)

PR #98, opened 2019-10-26
Minor french translation fixes

sephrat:
Typo fixes and rewordings.

Github newbie here - not sure I did the right thing when updating my fork from the original repo through a commit. Hope it doesn't screw up with this PR.


christianlupus:
This looks good. Thanks for contributing.

PR #105, opened 2019-11-11
App name from AppName Variable and some deleted templates that are not used

danielroehrig:
Before I continue cleaning up, I want to get this out of the way.


mrzapp:
Looks good to me :+1:

Issue #96, opened 2019-10-25
Incorrect Servings


paulcalabro:
I tried importing:

https://pinchofyum.com/spicy-shrimp-tacos-with-garlic-cilantro-lime-slaw

It's interpreting **Yield: 4 (2 small tacos each)** as **Servings: 42**.
I'm guessing this is a regex issue. What's the best way to debug this issue?

mrzapp:
@paulcalabro My suggestion would be to change [this](https://github.com/mrzapp/nextcloud-cookbook/blob/develop/lib/Service/RecipeService.php#L150) so that it makes use of regex to find the first uninterrupted string of numbers and parse it to an integer.

Would you be able to contribute this change, or should we handle it?

paulcalabro:
@mrzapp I can make that change. I'll take care of that a little bit later.

PR #109, opened 2019-11-26
Feature/incorrect servings#96

Teifun2:
Fixes Issue #96 

Added a simple Regex before the filter to select the first uninterrupted sequence of numbers as the correct number in the string.


mrzapp:
Thank you for the contribution, looks good to me :)

Issue #85, opened 2019-10-18
Add support for the "tool" keyword


Rinma:
Hi,

It would be nice to add the things I need to cook like the ingredients for eg. A mixer, A Pan, ...
When I want to cook something I know directly: Ok you need a pan, a spoon, and a mixer.

poltrian:
I Put These to the ingriediens

Marvin Dalheimer <notifications@github.com> schrieb am Fr., 18. Okt. 2019,
16:29:

mrzapp:
@Rinma we follow the [schema.org](https://schema.org/Recipe) format, which doesn't have a "utensils" field, but it does have a "cookingMethod" field. Should we add support for that?

Rinma:
What is with `tool`? This sounds like the right thing. 
When I get it right `Recipe` is a specialization of `HowTo` and therefore it can use `tool`, right?

mrzapp:
@Rinma you're right, `tool` is much more appropriate. I'll rename this issue to reflect that and place it on our development board.

Teifun2:
Does "Add support for..." indicate that the parser should also be able to handle this information? Or would this just mean allowing for manual input of the tool value?

mrzapp:
@Teifun2 should be both. I don't think anyone is working on this at the moment, though. Would you like to have a crack at it?

Teifun2:
I looked into this and i encountered a problem.
According to the [schema](https://schema.org/tool) a tool has two attributes `@type` and `name`. The [current parsing](https://github.com/mrzapp/nextcloud-cookbook/blob/09be169eac988b3937830e1f76dc737fb58fce84/lib/Controller/RecipeController.php#L121)  when editing a recipe is not able to handle this multi dimensional array. Or at least i don't know how to achieve that.

More specifically the following Form fields:
```
name=recipeIngredient[] val=1
name=recipeIngredient[] val=2
```

lead to the following json:
```
recipeIngredient[
    1,
    2
]
```

But what arrangement of Form fields would lead to:
```
recipeIngredient[
    [
        "@type": "a",
        "name": 1
    ],
    [
        "@type": "b",
        "name": 2
    ]
]
```

Teifun2:
I just noticed, that this is untrue... and i can just do it the same way as `recipeIngredient`

PR #112, opened 2019-11-30
Added Tool to Edit and View

Teifun2:
Adds the tool keyword to recipes as mentioned in #85 

The "tool" is already properly parsed if it is present in the json.

The Translation is no yet provided. I could add German Translation if someone tells me if i should use the `de` or the `de_DE` translation file.


mrzapp:
@christianlupus which one should @Teifun2 use?

christianlupus:
Sorry for my late reply. Both are relevant, one is the formal and one is the informal "you". So, both should be updated. I write this only for sake of completeness in case anyont stumbles opun this issue.

Issue #110, opened 2019-11-27
add "without category" in the "categorize selector" 


cloud2018:
To find recipes without category an additional item "without category" in the "categorze selector" dropdown would be helpful

In this way you could find recipes and add missing categories

mrzapp:
@Teifun2 I came across this while testing:

```
An exception occurred while executing 'SELECT "r".* FROM "oc_cookbook_keywords" "k" INNER JOIN "oc_cookbook_names" "r" ON k.recipe_id = r.recipe_id WHERE (k.user_id = ?) AND (k.name = "") ORDER BY "k"."name" ASC' with params ["jeppe"]: SQLSTATE[42601]: Syntax error: 7 ERROR: zero-length delimited identifier at or near """" LINE 1: ... r.recipe_id WHERE (k.user_id = $1) AND (k.name = "") ORDER ... 
```

Teifun2:
Oh, That's not good. I will look into that!

mrzapp:
@Teifun2 I just tested against a random library of categorised and uncategorised recipes. I'll dig a little deeper, maybe I can uncover this.

mrzapp:
@Teifun2 It seems to be a PostgreSQL specific issue, but still not sure how to work around it

mrzapp:
@Teifun2 It's this line: 

```php
->andWhere('k.name = ""')
```
I replaced it with a generated expression like this:

```php
->andWhere($qb->expr()->isNull('k.name'))
```

...but the logic doesn't return any results, no matter the data set.

You might want to approach this differently, perhaps like this (pseudo query, probably not correct):

`SELECT recipe_id,name FROM recipe_names WHERE recipe_id NOT IN(SELECT recipe_id FROM recipe_keywords)`

Teifun2:
@mrzapp I found out that Postgres seems to not like double quotes, so i swapped them around and it seemed to work!

`->andWhere('k.name = ""')` 

I will leave the Issue open until you can confirm that you also don't have any problems.

mrzapp:
Yep, works on Postgres now :+1:

cloud2018:
In Cookbook 0.7.5  it's not possible to find recipes without category, because "without category" is not displayed as folder:

In total there are 3 recipes, 2 with a category one recipe has no category

![cb1](https://user-images.githubusercontent.com/34692335/85797482-018a8f00-b73c-11ea-861c-63c42aa5536e.png)


**Versions**
Cookbook 0.7.5
[...]

PR #113, opened 2019-11-30
Added "Uncategorized" Category

Teifun2:
All Recipes without a single keyword are grouped under the "Uncategorized" word, as requested in #110.

The category is currently called "Uncategorized". Many Dictionaries seem to have a problem with this word although i could not find a clear indication that it does not exists / is not correct. Maybe we could think of changing it to something different?

I could add German Translation here as well.

### DIFF
--- a/lib/Controller/RecipeController.php
+++ b/lib/Controller/RecipeController.php
@@ -43,6 +43,8 @@ class RecipeController extends Controller
     {
         if (empty($_GET['keywords'])) {
             $recipes = $this->service->getAllRecipesInSearchIndex();
+        } elseif ($_GET['keywords'] === '-1,') {
+            $recipes = $this->service->getAllUncategorizedRecipesInSearchIndex();
         } else {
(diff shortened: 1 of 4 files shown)


mrzapp:
I think we just need to add the translation ourselves as you said.

Your approach is very respectful of the existing design pattern, I appreciate it 😊

I'll just go ahead and make you a collaborator, so you can push directly to `develop`, if you're making changes that don't require review.

Teifun2:
Thank you very much. 
But how do i know the changes that i make don't need a review? :D

mrzapp:
@Teifun2 that's mainly up to your personal judgement, as making you a collaborator places you within the circle of trust.

I would say that a bugfix or a minor enhancement usually doesn't need review, but a new feature usually does.

It can also just be if you would like feedback on your work for no particular reason :)

Teifun2:
@mrzapp Ok i see. Thank you for that trust!

Issue #108, opened 2019-11-19
Recipe editor: Pressing <Enter> with description field in focus opens file-selection window


stettberger:
When I edit an recipe and focus the description field, a press on the <Enter> key will open a modal window with an misleading title "Pfad zur Rezept-Sammlung" (Path to the recipe collection). The modal window is an file selection window.

While this behavior is useful for the picture field (if the modal window would have an appropriate title). It rather useless for the description field. My intention with pressing <Enter> was so save the recipe, therefore the current behavior was counter-intuitive for me and made me think that editing recipes was broken, since the save button is placed at the bottom of the recipe editor.

Issue #20, opened 2019-08-03
Pressing enter while in ingredients list should perform add


vozeldr:
While manually entering ingredients in the list, pressing enter should add a new row and set the focus to the new row. Currently, pressing enter in the ingredients list causes the value entered in the current row to be lost.

mrzapp:
@vozeldr I suppose that's because "enter" means "submit form". I'll happily welcome a pull request, if you've got an implementation in mind.

horga83:
I think this is an important feature.  It's very confusing for new users.  Took me a 3 or 4 tries before I clued in and my daughter and wife didn't get it until I told them.  They just thought the app was broken.

Hope you will consider this a bug and not a feature request.

mrzapp:
@horga83 I'll consider it a bug, sure

mrzapp:
Should be fixed now, I'll publish a new patch

horga83:
Many thanks, it's much appreciated, I forked the repo and started looking at the source but you know that takes awhile to get familiar with it.

PR #114, opened 2019-11-30
Feature/enter opens image selection#108

Teifun2:
This should fix the bug mentioned in #108 

While [searching](https://stackoverflow.com/questions/23569073/button-onclick-fires-when-enter-key-is-pressed-in-input-text-field-in-ie) i found out that it is expected behavior that a button in a form catches all "enter" calls and interprets them as "clicks" except if the button is specifically defined as a button.

The Filepicker name for the Recipe Image was wrong and was therefore double confusing.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -198,7 +198,7 @@ var Content = function (cookbook) {
         e.preventDefault();
 
         OC.dialogs.filepicker(
-            t(appName, 'Path to your recipe collection'),
+            t(appName, 'Path to your Recipe Image'),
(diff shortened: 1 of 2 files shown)


mrzapp:
Thanks, this looka good. The iaaue with "enter" you saw is the solution to #20

Teifun2:
Ah nice i didn't know that. It definitively also fixes #108 because that was what i was checking for :P

mrzapp:
Yep, I just closed it

Issue #118, opened 2019-12-06
Invalid argument supplied for foreach() at /var/www/nextcloud/apps/cookbook/templates/content/recipe.php#53


xsus95:
Invalid argument supplied for foreach() at /var/www/nextcloud/apps/cookbook/templates/content/recipe.php#53
Undefined index: tool at /var/www/nextcloud/apps/cookbook/templates/content/recipe.php#53

Since new update when opening different recipes.

Teifun2:
Could you provide me with the json of one Recipe that does not work? I'm so far unable to reproduce this error.

Teifun2:
The code referenced in this issue is a `foreach` over all `tools`. I first expected recipes to fail that have no tools defined, but at my end every possibility works. 
We could just wrap the `foreach` with an `is_array()` but i would like to now how this error comes to live.

xsus95:
[penne.txt](https://github.com/mrzapp/nextcloud-cookbook/files/3935469/penne.txt)

Please find the receipt attached to this message. While working fine running the previous update there seem to be issues with the current one.

Teifun2:
I'm still unable to reproduce this even with your provided recipe. 

Could you tell me more info about your server and your problem? 
I'm especially interested in your web server. Do you run nextcloud over an nginx or apache webserver?
When does the error appear? When editing / saving / loading the recipe?

If i would provide a test branch / would you be able to manually install the plugin?

Teifun2:
While searching around, i'm getting the feeling that this is possibly a issue due to the difference between nginx and apache. I see many reports of similar issues where the people always run nginx servers. It could be that Apache can handle empty foreach variables while nginx cannot.

If you can confirm that your running nginx that would be awesome. Otherwise i would be very confused ^^

Teifun2:
The code for Tools / Ingredients / Instructions is exactly the same. Have you ever created a recipe with no instructions or no ingredients? Does the same (slightly different) error appear?

xsus95:
Hey, the Error occures while click in a recipe to load. I use a nextcloudpi instance, so i think Apache. The recipe ist imported from https://www.chefkoch.de/ like all receipts who wont Work. But i know that i dont hat probs with the older version

Teifun2:
[penne_test.txt](https://github.com/mrzapp/nextcloud-cookbook/files/3949515/penne_test.txt)

Could you test this recipe for me?

Teifun2:
[...]

PR #119, opened 2019-12-11
Hotfix/foreach#118

Teifun2:
When adding the new schema word `tool` i forgot to implement the standardization for this keyword. This resulted in jsons that had no `too` key, if no tool was specified. Although this didn't result in errors for me it did for at least one user #118 . The `foreach` function seems in some php? versions be able to iterate over non existing variables (basically just ignore them). Some setups throw an error because the non existing variable is not an array.

I'm opening a pull request because i made changes to exisiting code which i'm unsure if they are correct. See in the comments.

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -231,6 +231,8 @@ class RecipeService
 
                 array_push($ingredients, $ingredient);
             }
+
+            $json['recipeIngredient'] = $ingredients;
         } else {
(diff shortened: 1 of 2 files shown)


christianlupus:
This could be due to the fact that the various server's PHPs have different configuration. You can make PHP more strict about non-existing variables and array members as far as I remember in the `php.ini` file. Just as an explanation for the behaviour. (The bug should be solved anyway!)

PR #123, opened 2019-12-19
Add missing translations strings for Spanish

agamez:


christianlupus:
Thank you for your contribution!

I am just on my way to update some translational files. Let me finish this then I can tell you if some minor modifications needs to be done for your PR.

PR #140, opened 2020-01-26
Interface overhaul

mcorteel:
Main changes:
* Add a home page
* Use global search
* Use a control bar like the Files or Deck app.

What do you think?

NOTE: I have a few other ideas, but I don't have time for them right now. Mainly:
* Remove the `<select>` and have the main navigation by tags, remove recipe list from app navigation
* Overhaul the recipe page.


mrzapp:
Like it, nice work! I'm putting this in the v0.6.0 branch, so we can mess with this some more before merging.

There are a few visual issues on the edit end content pages:

The image upload is a bit screwy:
image

Some padding missing here:
image

I support your other ideas too, and if you'd like any help with it, I'll be happy to assist.

mrzapp:
I've merged this with the v0.6.0 branch now, please make any future pull requests about this UI overhaul to that branch.

mcorteel:
@mrzapp What do you think about my idea for navigation ?
Screenshot_20200127_160014

It would require changing the client-side routing quite a lot to create new routes.
It would mainly mean prefixing the recipe routes (edit and show), but I think it would also be a good occasion to clean it up a little.

mrzapp:
@mcorteel I like it!

Instead of removing the keywords selector though, let's base these categories on the [recipeCategory](http://schema.org/recipeCategory) field, which we actually have yet to implement on the edit page.

Do I understand you correctly in that when clicking on a category or submitting a search query, the resulting recipes would then be displayed in the main view in a grid format? Because that would be awesome, I've been wanting to do that, but had no time for it yet.

I support making changes to the client-side routing as well, it's only been this way because we're still in early development. The v0.6.0 branch is open to any new ideas.

mcorteel:
Yes, that was my idea. Actually I would have done a list view, but a grid view with thumbnails is a great idea. Maybe both?

mrzapp:
How about we do a grid view for wide screens and a list view for narrow ones?

mrzapp:
@mcorteel, I've created a new issue [here](https://github.com/mrzapp/nextcloud-cookbook/issues/143), so we can track the progress on our project board. I've also invited you to be a collaborator.

Issue #142, opened 2020-01-27
Editing recipe title creates a new recipe


mcorteel:
* Create a recipe called "Rice", save it
* Edit the recipe and rename it "White rice"
There are now two recipes : "Rice" and "White rice".

That doesn't seem like expected behavior, but maybe I'm mistaken. If not, I can create a pull request to fix that when I find the time.

mrzapp:
That is a bug for sure. Are there 2 recipes in the search index or in the file system? Or both, perhaps?

mrzapp:
Closing, as this will be fixed in 0.6.0

dmbreakey:
This doesn't appear to be fixed; experiencing it in 0.6.2. At least I'm assuming this is a repeat of the same issue.

Some observations:

1. The extra entry appears to be a "phantom"; if I open either one, they both link to the same recipe, that has the correct title.
2. Checking the filesystem, the folder that the recipe is stored in was renamed to the new title, but the JSON file itself was not (it was still named using the original title). As a test, I tried renaming the JSON file to the expected (new) title, and it disappeared after a manual rescan. If I restore it back to the original filename, and rescan, it reappears, again with two entries.

For reference the one I attempted the rename with was the "Instant Pot Bacony Refried Beans" entry.

![Cookbook - Breakey Home - Google Chrome_565](https://user-images.githubusercontent.com/197063/77829685-44ab7680-70e9-11ea-936d-60e8ee302220.png)

dmbreakey:
Some more observations:

Just updated to 0.6.3, and I've noticed that the duplicate entry issue only appears to happen when viewing the Category it is attached to; the "All Recipes" view shows just one entry, as expected.

How can I revert back to 0.6.2 to verify whether it happened with that release?

Schlaubischlump:
@dmbreakey To my knowledge there is no easy way to downgrade a nextcloud app. You can download 0.6.2 from github and replace the changed files in the app folder manually. You can not copy the whole folder, because that will produce an NextCloud error each time you try to access the server.

mrzapp:
@dmbreakey could you get the latest from `master`, rescan your library and try this again?

dmbreakey:
Sure; I'll try to remember to do it tonight. Assuming I can just to a Git checkout, then copy the files over into apps/cookbook?

mrzapp:
@dmbreakey yep, that's pretty much it

dmbreakey:
[...]

PR #144, opened 2020-01-31
Router rewrite

mcorteel:
This work is very much in progress, but I wanted to push something to start  a discussion!

What was done:
- [x] Simplify routing:
  - `/cookbook#some/route` calls `/cookbook/some/route` and renders the response in the content wrapper
  - POST and PUT requests are sent to the current route (`POST /cookbook/{id}/edit` for `/cookbook#{id}/edit`)
- [x] Create new routes
  - Search (to display results from the search bar
  - Tag (to display all recipes in a tag)
  - All (to display all recipes)
  - Category (to display a category) (for later, see last point)

The main change is that the nextcloud app doesn't directly use the API anymore. My plan would be to move the API to a new set of routes (probably with a prefix) for external applications.

What remains to be done:
- [ ] Create new routes for the API (@mrzapp I'm not sure if the  API is actually used for now, but I would suggest that we put it behind a `/cookbook/api` prefix to separate it from the actual nextcloud app)
- [x] Reimplement all previous functionality using this new routing paradigm
  -  Recipe import (I would like to include that in the "Create recipe" page to remove clutter from the navigation)
  - Settings
  - Other things that I forgot
- [ ] Reimplement the page-specific JS in some way (like collections of input)
- [ ] Do the work on categories. For now, the navigation shows tags as an alternative, but like @mrzapp mentioned, real categories  would make more sense in the long term. This will probably be another pull request, though.

Comments are very welcome!


mcorteel:
OK, this is pretty much a complete proposal now! @mrzapp Looking forward to your suggestions :slightly_smiling_face:

mcorteel:
Note: This also fixes #142.

mrzapp:
@mcorteel Sorry, I've been super swamped with work the past month, but it's starting to clear up. These changes are truly phenomenal looking. Awesome work, sir!

About the feedback you requested:
- I agree that the API routes should be moved to `/cookbook/api`, is that the case in this complete pull request?
- Is there anything standing in the way of implementing the new categories? I really think the tags can just be used for the search index, or maybe a little tag cloud below the search box, if we want to get fancy.

mrzapp:
@mcorteel I've merged the changes into 0.6.0, in case that wasn't clear, BTW.

mcorteel:
* Yes, the routes have been moved to `/cookbook/api`
* I think categories should be pretty easy, but I haven't gotten to change anything in the database until now, so I will have to look into that.

mrzapp:
@mcorteel nice.

I've cleaned up a bunch and polished some of the UI changes, please pull the latest from 0.6.0 to see them.

There were a few regressions, such as the "import" not working since it was made into a generalised "submitForm" handler, so I had to roll that change back. I do like it better when I can tell just by the function name exactly what it does anyway, so I'd prefer to keep it like that, if you don't mind.

The timer was a stroke of genius, BTW. I've improved it to include seconds for the sake of visual feedback.

PR #157, opened 2020-03-14
Russian translation

MadFatVlad:
I have translated the strings to Russian language


christianlupus:
I think this looks good, although I cannot speak Russian. So, I will have to trust you with the language part. I will accept the PR here as only the newly generated language is affected.

Issue #46, opened 2019-09-01
Enhancement: use Transifex for translations


Lars1963:
While the nextcloud-community is very active in translating, I suggest you use nextclouds transifex-account for the translation of the app. This would give your app a amazing boost. 🚀

See: https://help.nextcloud.com/t/translation-knowledge-valid-for-the-entire-nextcloud-project-wiki/51550

The guys from nextcloud will more as happy in helping you to setup your repo for transifex. 😁👍

mrzapp:
@Lars as the time I get to spend on this app is very limited, I have actually been trying to contact the devs at NextCloud about moving this repo to their [organisation](https://github.com/nextcloud/), but I haven't gotten a response yet.

I bet the move to transifex would be a lot smoother if everybody didn't have to wait for me to have time to review pull requests. Do you have any idea how to move this along?

christianlupus:
I just looked into this issue (closing some old ones). I think [this documentation](https://docs.nextcloud.com/server/13.0.0/developer_manual/app/l10n.html#setup-of-the-transifex-sync) is useful. If you want, I can create the `.tx/config` file but the adding of the "collaborator" bug needs to be done by @mrzapp.

I do not want to avoid this transition to transifex but the currently existing translations might or might not get converted smoothly. I am unsure what will happen, it might be necessary to contact the translation team directly.

mrzapp:
@christianlupus not sure what you mean by "adding of the collaborator bug"?

christianlupus:
@mrzapp The repository has a set of collaborators. These can only be changed by the repository owner (or administrator). For Transifex to work, you have to configure some settings (I will prepare a PR soon if you like), add a certain user to the list of collaborators. And finally registering the app with the main CI system.

There is a daily schedule that revisits all registered repositories and synchronizes the translations to and from transifex system. The changed i10n location files are then updated within the repository by the bot. Therefore the bot needs write access to the repository. I think [this file](https://github.com/nextcloud/docker-ci/blob/master/translations/handleAppsTranslations.sh) is run on the repository by the bot.

christianlupus:
See here (from one of my repos):
![Screenshot_2020-03-16 christianlupus Test-PK-Pages](https://user-images.githubusercontent.com/8202922/76757533-07a9c080-6788-11ea-9282-74c72578b0c7.png)

mrzapp:
@christianlupus OK, let's give it a spin! Please prepare the PR at your leisure :)

[...]

PR #160, opened 2020-03-17
Added configuration for transifex

christianlupus:


christianlupus:
This is no longer useful. See #46.

PR #165, opened 2020-03-23
Fix cookTime, prepTime and totalTime are reset to 00:00

Schlaubischlump:
This fix should ignore the year, month and day values inside a time string and removes the debug `_test` values.


mrzapp:
Great, thanks!

PR #173, opened 2020-03-26
Migration to Vue.js front-end

### LINKED ISSUE #172: Front-end using Vue.js
Hello,

First of all, thank you for the great app! I started working on something similar a while ago, but I actually like your approach better. I could imagine myself using this app a lot.

There is one issue I would like to bring to your attention, though. For a good while now, Nextcloud seems to have been making a big move towards using Vue.js to handle the front-end. Their support for traditional app design i still good, but for the long term (and for consistency with the rest of the infrastructure), it would be good to consider using Vue.js in this app as well. The UI part of the app is still quite simple and the undertaking would not be overwhelming at this point. Having a couple of Nextcloud app projects using Vue.js myself, I can also tell you that the component-based design can really make things easier when you want to add dynamic features to you UI (such as the suggested recalculation for different amounts of servings). Nextcloud also supplies a lot of ready-to-use Vue.js components that have helped me in my earlier projects.

I'm a bit busy at the moment, but since I can consider putting the plans for my own recipe app to rest now, I could help you with the Vue.js rework in a couple of weeks. If you want to go down that road, that is.

Best regards,
Sam

PS: I have a couple of other ideas/suggestions for features for the app, such as subheaders in the ingredients section to improve readability in long ingredient lists, but I think I will return to those later. Depending on the direction you decide to take, I may be able to make pull requests for some of the features myself.

sam-19:
Resolves #172 (for linking).

### General idea

[Vue.js](https://vuejs.org/) is a javascript framework for UI development, similar in idea to React and Angular. Its main idea is to be lightweight to start with but incrementally adoptable, consisting of reusable components. My own experience with Vue has been that it is very easy to learn (compared to React for example) and very useful in projects with complex UIs. It does have some cons to, mainly the need to compile the javascript code (this initial commit includes an initiated node project in a subfolder `vue`) and to bundle it (I prefer WebPack for this). Increased complexity may also make it more difficult for people to commit to the project.

I want to open this pull request to discussion before you decide whether you want to adopt it or not. Moving to Vue would mean some fundamental changes to the app design. Although you don't _have_ to go all the way, it would be easiest to adopt the entire responsive design philosophy. Some of the changes would include:
- The index page would be the only synchronously loaded page left in the design; all other operations would be handled asynchronously. This means that the page would absolutely not work at all without javascript, but I think that is pretty much the norm nowadays.
- All templates would have to be converted into Vue components.
- Routing would be handled entirely by Vue. There is no need to change the routing scheme, but the handlers would have to be rewritten.
- Database transactions would be handled by an API, that returns the response as a JSON object. The existing database controllers and transaction services probably don't need any changes to adopt this.
- Translations should _probably_ be handled by Vue. It may be possible to dump the PHP translation into the DOM and parse it with javascript, but I have no experience in this. The i18n module in Vue is pretty comprehensive and mature in my opinion.

### Suggestion for initial component hierarchy

I like to have two root level components on the index page, one for navigation and one for content.
- Navigation - The left side navigation
  - Nextcloud has Vue components for pretty much anything you would need in the navigation, very easy to use and consistent with the rest of the platform.
- Recipe - The main recipe content. Following the current design of the app, the main page would have the following components:
  - RecipeImages - A component to show the recipe image; could be extended to include a slideshow of several images or whatever you want. Loaded asynchronously, so even large images + bad connection won't slow down displaying the main recipe.
  - RecipeHeader - A component that displays the name, description, source, and servings of the recipe.
  - RecipeTimer - Reusable component for each of the timers included in the recipe. Can be configured to use as a countdown (like the cooking time now has).
  - RecipeSection - Optional parent component for ingredients, instructions and other similar components (that may be added in the future). Could be useful if, for example, you want to add the possibility to collapse some sections by clicking the header.
  - RecipeIngredientList - List containing the recipe ingredient components
    - RecipeIngredient - A component for one individual recipe ingredient
  - RecipeInstructionList - List containing the recipe instruction components
    - RecipeInstruction - A component for one single step in the instructions

The same idea would go for the edit page:
- EditText - Component containing a text field row (title and input)
- EditImage - Component for the image input field row (title, input and browse button)
- EditTime - Component for a time field row (title and both inputs)
- EditKeywords - This could be changed to an auto-completing multiple select component (a Nextcloud Vue component)
- EditIngredient - A single ingredient row, could possible include drag and drop to rearrange (I think Nextcloud has a Vue component with this functionality, will have to check)
- EditInstruction - A single instruction row, same as above

I have not used breadcrumbs in my previous projects, but I have a feeling Nextcloud has a component ready for that as well.

Ideas, questions?


mrzapp:
@sam-19 I think your proposal sounds legit, but honestly you're the only developer with real Vue.js experience on this project that I am aware of, so the stage is yours on this one.

I will be happy to assist (and @mcorteel too, I'm sure) in any bug squashing, consultation and feedback along the way.

TheMBeat:
@sam-19 are you still thinking about porting the UI to Vue? I don't know Vue very well, but if there is time and I get help, I would like to participate.

sam-19:
Hey, @TheMBeat! Yes, I am. I initially said I would have time in a couple of weeks but that may have been a little optimistic. I'll try to get started with it during the end of this week. I'll take a top-down approach, probably starting with the recipe view page and continuing on one parent component at a time, whenever I can squeeze it in. I was debating whether to do it in javascript or typescript, but I guess Vue is challenge enough for most people and going with typescript in addition to it could be bit too off-putting for most.

I will do my best to document my progress here, so if you have any questions or suggestions, or want to push some changes of your own, just let me know :)

mrzapp:
@sam-19 bear in mind that some minor UI fixes will be made on the `master` branch, which need to be taken into account here.

Usually @mcorteel is the author on those, so it would be ideal if you guys could update each other as you go.

sam-19:
@mrzapp I see, the base branch of my pull request (v0.7) is already quite a bit behind of the master branch. Perhaps you could push the changes from master to the v0.7 as well so I could rebase my merge request to avoid any conflicts? Or do you have some other way in mind to manage the project branches?

mrzapp:
@sam-19 I've added you as a collaborator, feel free to merge in `master`. I imagine I might not know which bits to keep, if there are conflicts :)

sam-19:
Okay, chees @mrzapp! I will do that once it becomes relevant, right now I'm still working exclusively in the the 'vue' folder. I'll push my first changes soon. I've tried to explai as much as I can by commenting the code, so if @TheMBeat or someone else wants to jump aboard and contribute, reading some Vue introduction and following the comments should get you started.

By the way, Nextcloud seems to be removing their support for jQuery in the next major release and a lot of my ready-to-use code from previous projects relies on it. Do you have plans on shipping jQuery with your app or would you like to try and do without it completely? It will need some additional work if we don't want to use jQuery.

mrzapp:
@sam-19 I'm in favour of not using jQuery, and I'll be happy to assist with it.

I suggest you do your implementation the way you're most comfortable with initially, and then afterwards we can prune the code base together for jQuery references and replace them.

It's more important not to get stuck on initial implementation.

sam-19:
Okay, I have made an initial version that covers the recipe view timers, ingredients, tools and instructions as Vue components. Being only "half" of an app it includes some hacky workarounds like an interval loop to check if the app has been dismounted; see code comments for more info. These will be removed once the app handles the entire view. I have moved most if not all of the relevant javascript and CSS to the components and commented them out of the script and style files. Using scoped javascript and CSS has some benefits and some drawbacks. The obvious benefit is that they are "where you expect them to be" (that is, close to the components that they apply to). One drawback is that it introduces some overlap, like having to repeat the li styles in both ingredients and tools components.

I added one feature; subheaders in the ingredients. This is not a part of the schema.org standard, but it would make sense if you want to introduce support for markup in the app later. Many of the core Nextcloud apps support markup, so it could be the right way to go, at least for parts of the app, like instructions. So my subheader implementation uses "## " at the start of the ingredient name to denote a subheader (because that's what it means in markdown). Example: an ingredient named "## For decoration" will print a subheader "For decoration" instead of a regular ingredient. To align the subheaders with the ingredients, I had to increase the left padding a bit, to 1.25em. Let me know what you think of this.

It may take me a couple of days before I can get back to this, life is kind of busy at the moment. See how you like it, any feedback is welcome.

TheMBeat:
Hello @sam-19, a nice well documented delivery. I have read all the changes and comments and would like to have a look at it. Now I have the question how to create a dev version. I had thought about a docker container with Nextcloud in which I test the app. 

How do I compile the app?
Or is there a better solution to create a dev version?

mrzapp:
@sam-19 I think your solution for the ingredient group headers is just about the only way to go. As long as we don't add any non-standard field names to the JSON object, I'm fine with it :)

So, all in all, this is actually just markdown support for recipe fields, which I believe is a good idea, as we've discussed it before on this project.

@TheMBeat you should be able to just `docker-compose up --build -d` this thing, unless I overlooked something:

**docker-compose.yml**

```yaml
version: "3.4"

networks: 
    myserver--postgres: ~

services:
    # PostgreSQL
    postgres: 
        container_name: myserver--postgres
        environment: 
            - POSTGRES_PASSWORD=XXXXXXXXXXXXXXXXXXX
            - POSTGRES_USER=postgres
        image: postgres:latest
        restart: always
        volumes: 
            - "./postgres/init:/docker-entrypoint-initdb.d/"
            - "./postgres/data:/var/lib/postgresql/data"
            - "/etc/localtime:/etc/localtime:ro"
        networks: 
            - myserver--postgres

    # NextCloud
    nextcloud: 
        container_name: myserver--nextcloud
        depends_on: 
          - postgres
        image: nextcloud:latest
        restart: always
        ports:
            - "XXXX:80"
        volumes: 
            - "./nextcloud/data:/var/www/html/data"
            - "./nextcloud/config:/var/www/html/config"
            - "./nextcloud/themes:/var/www/html/themes"
            - "./nextcloud/custom_apps:/var/www/html/custom_apps"
            - "/etc/localtime:/etc/localtime:ro"
        networks: 
            - myserver--postgres
```

sam-19:
Hey! You can run the app by pulling it from the repository, @TheMBeat, since the compiled javascript file is included. Before merging into the main repository we have to either ignore the file or at the very least compile it with a terser to write it all on one line; right now the thousands and thousands of lines of compiled script are polluting the commits.

If you want to develop the Vue part, you need to have both Node and NPM installed. Browse into the /vue directory and in a console, run `npm run devel`. This will compile the project in development mode and watch the files, so it will automatically recompile if you change something. `npm run build` is for building production versions and is not yet fully configured, so you shouldn't use that command at this point. I find docker the best way to go for development. Just make sure you mount the apps in the custom_apps folder (like @mrzapp has in his example above) to avoid code integrity check errors.

I will have more time to use on this again next week, but feel free to ask me if you need help with something in the meantime!

sam-19:
Okay, the entire recipe view is converted to Vue now. It has some strange quirks related to only being part of an implementation, so don't worry if everything doesn't work like it should. I didn't find a way to make the breadcrumb action buttons look like you have in the current app, so that may still need some looking into. I also took the liberty of increasing the spacing between recipe description, servings and source a bit, and styled the description text in italics, to make them them visually more distinguishable. These changes are easily reverted if needed.

I haven't yet quite figured out where you get the _$ (recipe object) from, @mrzapp? Could you say if there is a way to fetch that object asynchronously as JSON? That would be the next step to move forward with this.

mcorteel:
@sam-19 As you can see in `MainController.php:176` versus `RecipeController.php:60`, some information is missing from the API, but:
* The setting to display the image when printing will be loaded with the app
* The image URL could probably be moved to `RecipeService::getRecipeById` anyway, couldn't it @mrzapp?

mrzapp:
@mcorteel I think the question @sam-19 was asking was which concrete API URL would they need to use to fetch the recipe JSON data. I'm actually not so sure about that since the latest changes in the router. How does that "resources" entry work?

mcorteel:
Oh, sorry. As explained [here](https://docs.nextcloud.com/server/18/developer_manual/app/requests/routes.html#registering-resources), the route will be `GET /api/recipes/{id}`

mcorteel:
Wow! This is really coming together nicely!

sam-19:
In fact, it has reached a state where it is kind-of working and has most of the functionality of the current app. There are a couple of issues I haven't been able to solve yet, like how to pass the current translation engine to the app. Nextcloud manuals suggest using `Vue.prototype.t = window.t` but in my limited testing it didn't work. In any case, if I do some final preparations, like a production build step, ignoring the scripts directory, and renaming the vue directory to src (as seems to be the convention with Nextcloud apps), we could consider merging it to the 0.7 dev branch. That way people can start working on other issues of the 0.7 milestone without having to worry about major conflicts.

There is one suggestion I would like to make to the filesystem/db API. Right now there are some inconsistent field names depending on your query. As and example, getting a single recipe it returns an object with "id" as the recipe ID, but getting all recipes or all category recipes returns a list of items with "recipe_id" as the ID. My first suggestion would be to simply use "id" everywhere (same for a couple of other field names). Also, since the Nextcloud API, for some reason, seems to automatically convert the PHP-style (or frankenstyle) object keys to json-style (cameCase) key names. So if there is some reason why the "recipe_id" (and maybe some other fields) need to be named uniquely, we should consider converting the keys to a json-style format (e.g. recipeId) in the controller. I already made this change in one controller method for `image_url -> imageUrl`.

What's your take on this, @mrzapp?

sam-19:
Oh, and on the subject of production builds, do you think we should completely ignore the scripts directory contents or should we commit a minified version of the script to the master repository? If we include the minified script, people can run the app by cloning the repository directly. This will make it easier for someone to develop other parts of the app, as they don't have to build the script first. It would introduce some complexity though, as someone who is doing changes to the javascript part would have to change the loaded script to the development version first, AND would have to change it back before committing to the repository. I'm not sure how Nextcloud handles loading the script files, it isn't very clearly stated in the documentation. If `script('cookbook', 'script')` actually prioritizes `js/script.min.js` over `js/script.js`, the developer could simply remove the minified version while he's testing changes, and then run the production build step before committing. In that case it would be a simple step of writing a check to to the production build to make sure the `script.min.js` exist.

mcorteel:
Sorry to bring this up, but do we really want to omit semicolons in JS?
I know there are justifications for that, but I'm personally against it.
Thoughts?

mcorteel:
> Oh, and on the subject of production builds, do you think we should completely ignore the scripts directory contents or should we commit a minified version of the script to the master repository?

With the only dependency being NPM (which is expected for Vue.js apps), I don't think we should bother too much with that. I would just add a "Development" section to the REAME.md with the two commands needed to start hacking: `npm i` and `npm run dev`.

sam-19:
> Sorry to bring this up, but do we really want to omit semicolons in JS?
> I know there are justifications for that, but I'm personally against it.
> Thoughts?

This is a valid question. In the past, I used semicolons with javascript as well. Nowadays most of the open source projects that I have worked with omit them. The main reason I stopped using them was that https://standardjs.com/rules-en.html recommended avoiding them. I don't follow all of the rules in that guide though. If we want more people to take part in developing tha JS side of the app, we're probably going to have to start using a linter anyways, and those can either automatically add the semicolons or remove them, so it's not a lot of trouble to switch the style. My personal preference is not use them, because they are unnecessary, not recommended by a few guides that I have read, and seem to be omitted in most modern javascript projects that I have come across.

On the same topic, I have changed the `self = this` in the original script to `$this = this`. As a python programmer I can appreciate the lure to use `self` as an alias, but almost every time I have seen other people referencing `this` for callbacks it has been as `$this`. So from my limited experience it seems to be the more common way.

sam-19:
It seems to be working pretty well in my testing. The settings pane in the navigation is still a work in progress, re-scanning the library and setting a recipe folder work, but the update interval and print image checkbox do not do much at this point. Also, it doesn't load the current values from the database yet. I managed to pass the Nextcloud default translation method to the App, but it's kind a of bandaid solution and needs to be checked by someone who's familiar with the way the translation system works. From what I can tell, it at least would need to accept an object of substitution values in addition to the translation string.

mrzapp:
@sam-19 the translation should work through a global JavaScript method, like in this example:

`t(appName, 'Could not set recipe update interval to {interval}', {interval: interval})`

Is that what you were looking for?

**About .$this vs .self**
I think both are pretty ancient practices, `.this` is all we really need. If we need to reference a class within a nested lambda function, all we need to do is declare it as an arrow `() => {}` instead of the traditional `function() {}`. Then `.this` will work as expected inside the lambda function. Also, unless guidelines explicitly prohibit it, I think we should move to `class` instead of the hacky `function` and `prototype` structure.

**With or without ;**
I think @sam-19's point is completely valid, and coming from a Python background, I can see why leaving out semicolons is preferred. However, like @mcorteel, I also don't agree with the style as proposed in the https://standardjs.com/rules-en.html. I could go on an on about why, but the only reason we really need is this: The NextCloud codebase uses semicolon, so apps written for it should too. It's always good practice to follow the core's conventions, whatever they are.

mcorteel:
@mrzapp You're taking words out of my mouth (or keyboard)!
I exactly agree with your arguments for semicolon and against `$this` and `self`, ES6 rules!

mrzapp:
@sam-19 by the way, just checked out the latest code, and hot damn is it snappy! Can't wait to finalise this version :smiley:

sam-19:
> 
> 
> @sam-19 the translation should work through a global JavaScript function called `t(appName, template, context)`, like in this example:
> 
> `t(appName, 'Could not set recipe update interval to {interval}', {interval: interval})`
> 
> Is that what you were looking for?

Yes, that's the one I'm talking about. Right now I made a function in main.js that fills in the appName automatically, but doesn't pass through the `, {interval: interval})` part, so that needs a bit more work.

> **About .$this vs .self**
> I think both are pretty ancient practices, `.this` is all we really need. If we need to reference a class instance within a nested lambda function, all we need to do is declare the lambda as an arrow `() => {}` instead of the traditional `function() {}`. Then `.this` will work as expected inside the lambda function.

I have no problems admitting that my javascript coding standards may be outdated :). I am not familiar with all (or even many) of the ES6 nuances, will have to look into it. It certainly sounds a lot better than the $this and self tricks!

> **With or without ;**
> I think @sam-19's point is completely valid, and coming from a Python background, I can see why leaving out semicolons is preferred. However, like @mcorteel, I also don't agree with the style as proposed in the https://standardjs.com/rules-en.html. I could go on and on about why, but the only reason we really need is this: The NextCloud codebase uses semicolon, so apps written for it should too. It's always good practice to follow the core's conventions, whatever they are.

I understand this reasoning, but would like to point out something else. In my limited sample (basically the first three listed and actively developed major apps in the Nexctcloud base repository, namely [Text](https://github.com/nextcloud/text), [Calendar](https://github.com/nextcloud/calendar) and [Mail](https://github.com/nextcloud/mail)), none of these apps use semicolons in their javascript code. So, instead of making the decision based on Nextcloud codebase, we could also consider the prevailing app development standards, as we are more likely to attract the other app developers than Nextcloud codebase developers to help with the project. But like I said, with modern linters, it's a matter of flipping a switch so to say, to change from one style to another.

All in all, the Vue rework is close to done, just some minor tweaks left for the settings and maybe some other parts that I have missed.

sam-19:
Okay, @mrzapp and @mcorteel, the current iteration has passed all my tests so far. I tidied it up a bit, removing unnecessary files etc. Some issues remain, but I don't think I should start sorting these on my own:
- The app does not fetch values for the settings pane, because there was no pre-existing API call for that and I didn't feel like messing around with it any more than necessary (I already repurposed one or two controller methods for the app). This should be left for the routes and API overhaul.
- Search page doesn't exist yet. For some reason the `disabled` flag isn't working on the action button on the breadcrumb row, so maybe it should just be removed for now. Also, for those who have a lot of recipes, maybe loading a set amount at a time with the page navigation after the index page header in breadcrumbs would be better, with the search page link placed somewhere else.
- The auxiliary method for the translation engine (`Vue.prototype.t`) still doesn't pass through any extra parameters.
- I couldn't quite figure out what the automatic update in the settings menu does, so it does nothing at the moment...
- Most action buttons are missing tooltips at this point.

As a side note, I noticed the width of the total time field in recipe view is less than the width of the other two time fields. I tried to fix it but couldn't figure out the root cause. It probably has something to do with the flex display, but I'm not terribly familiar with it.

mcorteel:
I can't run the app anymore. Just using `npm run dev` worked like a charm before, but now I don't know where the assets are and I get a 404 error when fetching the `vue.js` file.
@sam-19 Any ideas?

sam-19:
The 404 was to be expected, as I have ignored the compiled script file from the repository (as discussed above). The fact that `npm run dev` worked like a charm before is a bit perplexing though, because I only renamed the command to that last night; it was called `npm run devel` before this. Perhaps you are trying to run it in the wrong directory? Browse into `/src`, run `npm i` and then `npm run dev`. It worked for me at least. The previous `vue` directory has been renamed `src` in preparation for merging the branch (also as stated above), but maybe I should have made this more clear, sorry. The previously mentioned `npm run build` is working also now, but the resulting script is still too big and will require some chunking before this can be released in the AppStore.

mcorteel:
Very nice! I was running the command from the app root... And I don't know why it kind of worked before...

I don't have much time to test, but it mostly works very well! Thank you for this awesome work @sam-19!

sam-19:
A reasonable option would be to actually move package.json to project root, so one could run the commands there. Storing it inside /src is a habit from projects where the Vue GUI is just a small part. In this project it has quite a major role. It would also remove the need for one dependency (filemanager-webpack-plugin).

mrzapp:
@sam-19 I agree, we should just move it to the root, and as @mcorteel said, we can just merge this pull request with 0.7, as there aren't too many other changes in that release.

sam-19:
Okay, @mrzapp, it is ready for merging. I also added some initial config for chunking, if you decide to do that later (and I think you should).

PR #179, opened 2020-03-28
Hide zero times

jvasile:
When importing and editing recipes, if you end up with 0 values for
the cook or prep times, the recipe page will show big boxes displaying
unhelpful 0 values.  This commit omits those boxes as needed.

I offer this code under the terms of the AGPLv3 or any later version.


mrzapp:
Makes sense, thank you :)

PR #181, opened 2020-03-29
Make browser cache recipe images until changed

daniel-k:
I noticed that my browser reloads all the thumbnails everytime when the home page of the Cookbook app is loaded. After some investigation I found out, that a dynamic `t=TIMESTAMP` parameter is added to each image URL. I guess this is to explicitly prevent the browser from caching the images which would be a problem when the user changes a recipe image.

This PR implements a PoC for setting the `t=` parameter not to the request time but to the modification time of the requested image as well as the header `Cache-Control: public, max-age=604800`. This way, the browser will be able to cache all images but still reloads them when the user changes the recipe image.

Please bear with me concerning the implementation. I usually don't write PHP code and couldn't figure out where the appropriate place for the `imageModificationTime()` function would be. In order to produce a working PoC, I simply duplicated the function which is really bad practice I admit. Feel free to shuffle around my code as you desire :)

And btw: thank you for this awesome Nextcloud app :upside_down_face:

### DIFF
--- a/lib/Controller/MainController.php
+++ b/lib/Controller/MainController.php
@@ -66,6 +66,22 @@ class MainController extends Controller
         return new DataResponse($keywords, 200, ['Content-Type' => 'application/json']);
     }
 
+    public function imageModificationTime($id, $size)
+    {
+        $path_candidates = [
+            $this->service->getRecipeImageFileByFolderId($id, $size),
+            dirname(__FILE__) . '/../../img/recipe-' . $size . '.jpg'
+        ];
+
+        foreach($path_candidates as $path) {
+            if(file_exists($path)) {
+                return filemtime($path);
+            }
+        }
+
+        return -1;
+    }
+
     /**
      * @NoAdminRequired
      * @NoCSRFRequired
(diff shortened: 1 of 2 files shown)


mrzapp:
Based on your code, I've tried just setting the `max-age` and remove the `t` parameter from the image URL, and that seems to do the trick, at least after a reload.

Can you check out the latest `master` and verify that it works for you?

daniel-k:
I'll check this evening, thanks for the quick response!

Without running the code, I think we still need to stick to the `t=` parameter. Otherwise, the browser would keep using the cached image even after the user has changed the recipe image. Could you also test that?

mrzapp:
@daniel-k I am getting fresh images after a soft reload, but that might still not be great UX

daniel-k:
> I am getting fresh images after a soft reload, but that might still not be great UX

Do you mean, that after changing the recipe image you see the old image until you do a soft reload (I guess pressing F5)? That's what I expected and this is why I changes the `t=` parameter to be the modification time of the recipe image. But you didn't include it in you commit on master.

daniel-k:
> Do you mean, that after changing the recipe image you see the old image until you do a soft reload (I guess pressing F5)?

I can confirm this behaviour with the current `master` branch.

mrzapp:
@daniel-k I see! That's smart, I didn't really get it the first time I looked at it.

I've implemented your idea, but using the recipe file mtime, since it's more general and fits neatly into the `dateModified` field of schema.org. Also, the image is very unlikely to have a modification time more recent than the recipe file.

Please check out the latest again and see if works as you intended.

daniel-k:
Nice, this looks like a much simpler way even without hitting the filesystem. :+1: I like the idea of taking the recipe modification time! While this will make the images reload after a recipe update even if the image wasn't updated, I think this is a reasonable tradeoff here! Will try this evening.

daniel-k:
Works like a charm! Thank you very much :upside_down_face:

daniel-k:
Feel free to close this PR whenever you consider it done :wink:

Issue #161, opened 2020-03-17
Print Recipe with Image


phantomdez:
I cannot seem to find an option for this and maybe it hasn't been added as of yet but when I try to print the recipie it doesn't add an image.  Am I missing an option or has this been done by design?  My wife would like to include a image for her physical copies that she keeps in here Recipe binder.

Thanks

mcorteel:
Yes, it is by design. I guess an option would make sense. @mrzapp?

mrzapp:
@mcorteel if so, should we make the print button a dropdown menu with 2 options, with and without pictures?

mcorteel:
It's a possibility, but I think that it goes against Nextcloud UI guidelines. I would have implemented it as a global option, but I guess it's not great either.
Maybe we should ask the Nextcloud usability guys? I'm not sure, but I guess @jancborchardt would be the way to go. Any comment?

jancborchardt:
I’m glad to help, but could I get some more context, maybe in the form of screenshots? :)

@mcorteel you mention it is leaving out images by design, why is that? If the recipe has an image set, I would expect it to be printed as well. Especially for recipes, an image makes it much nicer.

mrzapp:
@jancborchardt it was just what we considered common sense at the time, saving ink and all that. But it wasn't intended to be restrictive, we should definitely add it as an option, if it's wanted.

phantomdez:
I would definitely vote to have it as an option.  I brought this up as I thought I was missing a setting but since I see it isn't [Thank you all for your responses] I would vote to add it as an option.

cloud2018:
In Cookbook 0.7.6 the text in Settings is "Print image with recipe". I think "Print recipe with image" is more popular. 
![cb20202627](https://user-images.githubusercontent.com/34692335/85921274-71536380-b87b-11ea-82ab-3a7e500dccec.png)

mrzapp:
@cloud2018 patches welcome here too :)

PR #194, opened 2020-04-12
Improve input groups and use one for update interval setting hint

mcorteel:
In order to add more settings for #161, I wanted to improve the look of current settings, so I used a tooltip to display the last update time.

image

And to do that, I created a shared `.input-group` class with an `.input-group-addon`, (based on bootstrap3 vocabulary) and replaced the current `.list-item-tools`.

I also removed the `max-height` on buttons in this case, so that they fit textarea height, which looks better IMO (see screenshot below). But I guess that's not for me to decide ultimately :)
Also, I fixed button borders on hover.

image

Looking forward to your comments!


mrzapp:
The max-height was intentional, as I consider it best practice to keep button dimensions concise. We could try aligning them vertically on textareas, if that would be less of an eye sore?

mcorteel:
Another suggestion would be this:
image

mrzapp:
@mcorteel right, that's more like it. I'd prefer the buttons to be consistently "clicky", though, so how about this?

image

mcorteel:
That's cool, I like it. I'll try to make this work.

mcorteel:
There it is!
image

mrzapp:
Beautiful!

PR #198, opened 2020-04-13
fix #196 - use content attribute if it exists for ingredient element

### LINKED ISSUE #196: hidden text should be ignored when downloading recipes
On BBC Good Food, there are divs that show some text when you hover over some ingredients.

Example URL: https://www.bbcgoodfood.com/recipes/roast-lamb-studded-rosemary-garlic

This text that is hidden should not be included in the text of the ingredients list. For this specific site, there are two possible solutions:

1. There's a `content` attribute on the `li` for the ingredient, that contains the correct text
2. The unwanted text in the hover is contained in a div with `aria-hidden="true"` and so could be ignored as screenreaders should.

crazystick:
this will use the content attribute as suggested

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -474,7 +474,11 @@ class RecipeService
                         
                         if(!isset($json[$prop]) || !is_array($json[$prop])) { $json[$prop] = []; }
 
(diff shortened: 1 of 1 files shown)


mrzapp:
Nice, thanks for the contribution!

Issue #176, opened 2020-03-26
Ugly Picture


hitam4450:
Hello,

i just uploaded a sample of the picture added to the recipe !
it looks more then like a  banner then less  like a picture !
![cookbook-pic](https://user-images.githubusercontent.com/9738727/77672381-2169b580-6f89-11ea-8b2e-fd77c9b8c0b8.jpg)

Schlaubischlump:
Maybe showing a fullscreen preview when clicking on the image would be a good idea

hitam4450:
Thanks, I did but nothing happened !

Schlaubischlump:
You misunderstood. The feature does not yet exist. It was just a suggestion on how to keep the existing design and still be able to view the images in full size.

mrzapp:
@hitam4450 I don't think we're cropping the picture within the app. Can you check the image on disk, if it's cropped like a banner? If so, then it's the source (the website you got the recipe from) that decided to supply that image.

@Schlaubischlump the logic for fetching the largest image available is to loop through all image sources, dig out any numbers in the path, add them together and pick the source line with the highest number. It's not great, but I haven't been able to think of anything better, other than maybe letting the user pick one manually. Thoughts?

mcorteel:
@mrzapp The image __is__ cropped (using CSS positionning and `overflow: hidden`) to avoid gigantic images taking too much screen space.
I really like @Schlaubischlump's idea to allow clicking on the image to display a full size version:
* Either display it in some kind of modal view
* Or uncrop it on click.

mrzapp:
@mcorteel alright, then I think we could just uncrop it on click, that's the least troublesome overall. You on that?

hitam4450:
Load it as full size picture, e.g. one click present it as banner ( like now) double click as another format .... ideas are unlimited ...   :-)

PR #203, opened 2020-04-16
Implement setting to print image

### LINKED ISSUE #161: Print Recipe with Image

mcorteel:
Fixes #161.

Adds a checkbox in the settings to choose if the image should be printed with the recipe. The image is printed as is (see #176).

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -529,7 +529,7 @@
     #app-navigation,
     #controls,
     .recipe-toolbar,
-	#app-content header,
+	#app-content header:not(.printable),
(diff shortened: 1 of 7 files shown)


mrzapp:
Makes sense to implement it as a setting :+1:

Issue #183, opened 2020-04-01
Feature request: Clear URL field after import


dm7ds:
Hello,

automatic clearing the URL field after (successful) import of a recipe would be handy.

Thanks!
Dirk

mrzapp:
You're right, I've thought about it too :)

PR #223, opened 2020-05-16
Reload recipes after rescan

### LINKED ISSUE #168: Reload the recipe list after rescanning the library
Not a big issue, just a minor inconvenience. If this "bug" is not deemed worth to be fixed, just close this issue.

Steps to reproduce: 

1. Manually upload / delete a recipe from the folder structure (not in the cookbook app)
2. Open the cookbook app and rescan the library
3. The "All recipes" list does not update. You have to manually click on the "All recipes" sidebar entry to reload the view.

sam-19:
Resolves #168 and #183 (merge my other pull request first)

This performs just a crude refresh of the router view. For a more sophisticated solution (like only reloading the recipes and not the whole view, which would avoid the view flashing empty for a split second) some more fine-grained checking would be needed.

### DIFF
--- a/src/components/AppNavi.vue
+++ b/src/components/AppNavi.vue
@@ -287,6 +287,10 @@ export default {
                 $this.scanningLibrary = false
                 console.log("Library reindexing complete")
                 $this.getCategories()
+                if (['search', 'recipe', 'edit'].indexOf($this.$store.state.page) > -1) {
+                    // This refreshes the current router view in case items in it changed during reindex
+                    $this.$router.go()
+                }
             }).fail(function (jqXHR, textStatus, errorThrown) {
                 deferred.reject(new Error(jqXHR.responseText))
                 $this.scanningLibrary = false
--- a/src/components/AppNavi.vue
+++ b/src/components/AppNavi.vue
@@ -287,7 +287,7 @@ export default {
                 $this.scanningLibrary = false
                 console.log("Library reindexing complete")
                 $this.getCategories()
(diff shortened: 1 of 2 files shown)


sam-19:
Actually, there is a potential problem with this solution. Depending on how the update interval is supposed to work, this could potentially reload a recipe that a user is editing, reverting all fields to their original values. So this will need some more thinking, if the update interval is meant to automatically reindex the library.

sam-19:
In fact, now that I think about it, this opens a whole can of worms. When a user begins editing a recipe, does the app lock that particular folder? If not, some way to resolve version conflicts has to be set up. I will remove edit from the views that are automatically refreshed for now. To think about it, I will also remove basic recipe view; after all, if you have the recipe open when it is deleted, that gives you a last chance to save the data in case deleting the recipe was an accident.

mrzapp:
@sam-19 so should we park this pull request for now?

mrzapp:
@sam-19 folders aren't locked as far as the API goes, it just writes files with whatever content is submitted in the "edit" form. 

I honestly think we don't have to auto refresh the main content view, just the sidebar. Let's not overcomplicate things :)

sam-19:
Right now it only refreshes the view if the user is on the index page or search page (i.e. there are likely many results on the screen), which I think the original issue was about. I could try to bake in a check for the search page, where it only refreshes the view if it actually has changed. Another possibility would be to show a message like "Results in this view have changed" and offer a reload button for the user.

mrzapp:
@sam-19 I think it's fine to reload them automatically, but just exclude the edit and recipe views. Maybe that's what you already did, and I'm just slow 😅

PR #255, opened 2020-06-11
Fix when json object is an array

Eothred:
I found that on allrecipes.com the last element of the array is the actual recipe object.

This fix searches through all elements of the json object found, if it is an array. Previous solution was to just pick the first object, which in the particular case would get a BreadCrumb object. If no recipe object is found, the last element is selected (instead of first) which should have no impact as the next check will say that it (still) isn't a recipe object and continue the outside for loop.

I checked that this worked on my installation, running NextCloud 18. Still I'm new here so please verify before merging!

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -448,7 +448,14 @@ class RecipeService
 
             // Check if json is an array for some reason
             if($json && isset($json[0])) {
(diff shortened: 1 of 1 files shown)


Eothred:
This is the same as #256 , you might want to consider that solution instead

Issue #241, opened 2020-06-08
Dutch translation is missing in 0.7.1 All labels are in English


RobRobM:
All Dutch translations are gone, every label is in English.

Ps. Thanks for all the hard work. I love the Cookbook app!

mrzapp:
Dankjewel @RobRobM :)

Transifex, the translation service we're using, should automatically translate strings with every update we do, but since moving to Vue.js, it seems we've run into some trouble.

@christianlupus @sam-19 do you have any ideas?

christianlupus:
The translations are still there in the `l10n` folder. So the transifex connection is working fine as far as I can tell. Now I am unsure about the effects of Vue.js on the import/localization.

I just checked the problem on my installation (German by default). On an old installation 0.6.5 everything was ok. On the 0.7.2 I can confirm the texts are mostly all back to English.

I have no experience in Vue.js. How does it work? I just looked (as an example) into [this line](https://github.com/mrzapp/nextcloud-cookbook/blob/13e6ee41f9674d770cda84dd05f109d925d32223/src/components/RecipeEdit.vue#L3) I do not understand at the moment how this is parsed. Is it going to be interpreted as HTML? Seems strange for me, unfortunately. The important part is that the function `t(...)` is evaluated. I fear this is not happening here at the right moment. I fear you need more in-depth knowledge of vue to get there.

sam-19:
I think the shortcut I had to take to pass the translation function to Vue may be the cause. This is actually why I wanted someone with experience with the translation engine to check it out first :) However, it's not a big deal to fix (actually a batch find and replace), assuming that is the cause. I will take a look at it and post a pull request later this week, but I will need someone else to test it.

hitam4450:
> The translations are still there in the `l10n` folder. So the transifex connection is working fine as far as I can tell. Now I am unsure about the effects of Vue.js on the import/localization.
> 
> I just checked the problem on my installation (German by default). On an old installation 0.6.5 everything was ok. On the 0.7.2 I can confirm the texts are mostly all back to English.
> 
[...]

PR #258, opened 2020-06-12
Re-enable string translations via NC's default i10n function

sam-19:
A possible fix for #241. There are probably some stray translations that I missed, so additional patches may be needed. Also, I only have English language installed on my Nextcloud, so I need someone else to test if this actually does what it should. It didn't break anything for me but please, test in a controlled environment :)


sam-19:
PS: This includes hiding the search button in my previous pull request. Let me know if you don't like that and I will remove it from this one as well.

PR #261, opened 2020-06-13
Add an .l10nignore fire

sam-19:
See https://docs.nextcloud.com/server/15/developer_manual/app/view/l10n.html#translation-tool
This file is not really necessary at the moment since that folder is ignored by git, but I imagine it would be safe to already have for possible release versions that include compiled files in the js folder. I only wonder if Transifex would have a problem with a non-existing folder in the ignore file. This is certainly not the root of the problem in #241, so it's up to you guys if you want this or not.


mrzapp:
Let's await #241 before we move on this one

PR #263, opened 2020-06-14
Toggle navigation on narrow screens

### LINKED ISSUE #260: No navigation menu on the left on desktop and mobile version, all browsers.
**Description**
The navigation menu on the left side of the app is missing when viewing cookbook in the mobile view or desktop view when recipe takes up the majority of the window. Before when clicking on the 3 dashes (left top) the navigation menu would open showing "create recipe" "recipe url" and the categories. Now this is only visible when the window is big enough to display navigation menu and recipe. Now when I click the 3 dashes, I navigate to all recipe's
 
**Reproduction**
1. Go to cookbook app on mobile browser or desktop browser 
(2. Navigate to any recipe)
3. On desktop reduce with to only show recipe and navigation menu dissapears 
3. click on 3 dashes (top left)
4. See all recipes instead of navigation menu

**Expected behavior**

**Screenshots**

**Browser**
Which browser are you using?

Update: navigating to recipe to reproduce problem not necessary.

sam-19:
A quick and dirty hack to fix #260. Doesn't look pretty and may not work on some edge cases, but this should be replaced by a standardized solution later in any case.


mrzapp:
It's better than not having it, for sure :) Thank you

PR #295, opened 2020-08-12
Create .gitkeep

rakekniven:
Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>


christianlupus:
May I ask why a `.gitkeep` file is needed? There is already a `.dummy` file that will avoid an empty folder been deleted by git if all translations would have been removed (which is quite uncommon).

rakekniven:
> May I ask why a `.gitkeep` file is needed? There is already a `.dummy` file that will avoid an empty folder been deleted by git if all translations would have been removed (which is quite uncommon).

See the docs. The Nextcloud project want keep the same structure for all repos. For sure we can rename the dummy file as well.

rakekniven:
Should I modify my PR?

christianlupus:
I just did so. Will merge directly.

Issue #218, opened 2020-05-08
Optimise database queries


thenameiwantedwastaken:
Every time I try to access my recipes page I get a 504 timeout. If I refresh the page it works okay. All other pages of my nextcloud implementation work okay, so it seems there is something about the recipe app that is simply too slow. I am running nextcloud via docker / nginx. I don't think it's a problem with my specific setup of nextcloud as all other aspects of nextcloud appear quickly.

Navigating within the recipes is fine once the recipe page has shown... but just the initial view is too slow.

mrzapp:
I am using the same setup as you, Docker with NGINX, and I've never observed this behaviour. You're also the first to open an issue like this, so I assume there is something else going on.

I can only imagine a timeout would occur if the filesystem is being slow for some reason. Are your recipes stored on a network mount that might not respond at times?

Also, if you can find some relevant log messages, it would help shed some light on the situation.

thenameiwantedwastaken:
Recipes are stored in the same folder structure as all the other nextcloud files. I can browse the recipe images and json files without delay via the nextcloud folder interface. However, if I actually try to open a json file to preview it in the nextcloud interface, I just get a spinning circle. This happens to all json files, not just recipes.

christianlupus:
Do you mazbe have a plugin installed in NC that tries to start up and open the JSON?

mrzapp:
@thenameiwantedwastaken are you perhaps using an older NC version than 18?

dietrichmd:
I too have this issue.  When I load cookbook, I get 504 error..upon reload, it loads fine.

Looking at the logs, it looks like a php memory error

 [proxy_fcgi:error] [pid 17264] [client 172.69.71.97:63198] AH01071: Got error 'PHP message: PHP Fatal error:  Allowed memory size of 2097152 bytes exhausted (tried to allocate 4096 bytes) in /home/drive.braeswood.pl/public_html/3rdparty/composer/autoload_static.php on line 446\n', referer: https://drive.braeswood.tk/index.php/apps/cookbook/

Virtual Mem is set per nextcloud instructions and this instance is allocated 512M for PHP-FPM.

Included apps in nextcloud:  https://i.imgur.com/fuIsXW7.png

In addition to the 504, occasionally I'll click a category and cookbook will just sit there like the link doesn't even exist.  A reload will fix this behavior as well.

Thanks

stenosis:
[...]

PR #297, opened 2020-08-14
Rework the database requests for performance

christianlupus:
As discussed in #218 the database is used currently in a very bad way. This PR tries to optimize a bit of these accesses.

I want to be honest: This is more of a hack an less a real implementation in good manner. I found quite some issues in the code, that I will open soon document. There are yet some edges and `TODO`s in the code I have to iron out.
If someone wants to have a look anyways and give some feedback or help with the issue, you are welcome. If you wanted to test it out, it should work. However, I suggest to have a backup of you JSON files :smiley:. I tried some cases here but for sure not all possible (edge) cases.

# Open points regarding this PR
- [ ] Add option to reindex tables from scratch (GUI plus backend)
- [ ] Debugging needed in case of bugs? -> export of tables useful for bug reports?
- [ ] Option: Abort reindexing after fixed number of recipes/time to avoid timeout and start on next invocation again
- [ ] Test code
- [ ] Potentially enhance readout of keywords from DB


christianlupus:
Before processing further with any optimizations, I wanted to ask for your opinion in these changes.

I will not review them myself, as I think - of course :grinning:  - that my changes are the holy grail. To get a more critical feedback, I am asking you to look into this PR so far.
I would convert the open tasks into issues (as most are merely minor feature requests) manually afterwards.

christianlupus:
I merge the changes here to get them into the main development branch. As only the DB is affected, we can always return to the old behavior without data loss.

PR #318, opened 2020-09-16
Filter and search recipes on the main page

### LINKED ISSUE #243: Search is not Working Page not Found
Hello,

at both 7.1 & 7.2 search is not working. "Page not found"

sam-19:
A filter and search feature for the main page. Both fields are found under a three dot menu on the breadcrumbs row. The features pretty much do what they say; filter just filters the results already on the screen, search runs a new query to the database. As to why there are two functions: the three dot menu requires at least two elements and I could not find any way to display just one field on the breadcrumbs row. If having two fields is too confusing, the feature has to be redesigned (the field moved to the left side panel, for example).

I'll wait for your for feedback.

Fixes #243


lib/Controller/MainController.php
```diff
@@ -119,7 +119,7 @@ public function search($query)
 			$recipes = $this->service->findRecipesInSearchIndex($query);
 
 			foreach ($recipes as $i => $recipe) {
-                $recipes[$i]['image_url'] = $this->urlGenerator->linkToRoute(
+                $recipes[$i]['imageUrl'] = $this->urlGenerator->linkToRoute(
```
christianlupus, line 122: I am a bit surprised about this change. Could you have mixed up with another bug fixure?

appinfo/routes.php
```diff
@@ -13,8 +13,6 @@
 	   ['name' => 'main#home', 'url' => '/home', 'verb' => 'GET'],
 	   ['name' => 'main#keywords', 'url' => '/keywords', 'verb' => 'GET'],
 	   ['name' => 'main#categories', 'url' => '/categories', 'verb' => 'GET'],
-	   ['name' => 'main#category', 'url' => '/category/{category}', 'verb' => 'GET'],
```
christianlupus, line 16: Is this really no more needed? I just found [this line](https://github.com/nextcloud/cookbook/blob/master/src/router/index.js#L28) but I am not sure if this is related (not JS/VUE expert). Just the red warn lamp when changing the API.

christianlupus:
Additionally, you might want to amend the latest commit and sign it off (by `git commit -s`).

Regarding the pure VUE part I am not able to help much at the moment. I would have to check out the version and test it manually. Unfortunately, I have no dev environment at hand at the moment. I might have a look next week.

lib/Controller/MainController.php
sam-19: I have actually explained this before to someone, but cannot find my comment anywhere (which is odd). In any case, I changed the front-end to use imageUrl instead of image_url at some point, so this change was necessary. I still can't remember why I did it, but I have a feeling it was to fix some bug. Or maybe I just wanted to move the JSON API to use more JSON-like attribute names, but didn't finish the job. I admit there is some discrepancy in the attribute naming and it should be resolved by either (a) reverting the imageUrl back to PHP style image_url (a step in the wrong direction, in my opinion) or (b) changing the rest of the JSON API attribute names to camelCase (would require extensive work and testing). In any case, I feel this is something to address in an issue of its own.

appinfo/routes.php
sam-19: The API is, at the moment, a mix of the old PHP routes (which are reachable only through manually writing them on the address line) and the new JSON API routes. I have not just gone ahead and pruned it because I am not familiar with the back-end and don't want to remove anything that is actually still in use. Unless the goal is to maintain both the Vue front-end and a backup synchronous front-end for people who cannot or refuse to use javascript, the obsolete routes should be removed at some point. Maintaining two front-ends is too big an undertaking for me at least, so unless there is someone else on the team to maintain the PHP template side, I suggest we get rid of it sooner rather than later. Again, this should probably have its own issue.

And yes, the line you spotted in the Vue router does replace the line that was removed.

sam-19:
> Additionally, you might want to amend the latest commit and sign it off (by `git commit -s`).

I don't know how this can be missing, possibly something has gone wrong with my VSCode settings. I'm away from my dev environment as well, but will resolve this next week.

lib/Controller/MainController.php
christianlupus: You are rigght. I just wanted to make sure this is not used anywhere else.

appinfo/routes.php
christianlupus: In fact, you moved the routes to the `/api` prefix [here](https://github.com/nextcloud/cookbook/pull/318/files#diff-0e033fd8797499a87a6bf2ffccef5c3cR28-R29). This is perfectly fine for me. I was just curious why in the vue router the prefix is not mentioned.

christianlupus:
Ok, I started by doing a local merge with my most up-to-date local branch.

Regarding the filtering I see the following bug: If you apply a filter, close the small overlay window (you created with the filter and search) and reopen it, the filter term is empty. So to clear the filter you have to type something and remove it afterwards.

The search seems not to be working. I am not (yet) aware, where the problem lies. I get always an empty result list from the backend. Further debugging is needed.

christianlupus:
I found the cause for the empty answer from the backend and corrected things accordingly.

sam-19:
> Regarding the filtering I see the following bug: If you apply a filter, close the small overlay window (you created with the filter and search) and reopen it, the filter term is empty. So to clear the filter you have to type something and remove it afterwards.

This should be fixed now.

christianlupus:
@sam-19 I made some manual changes in order to make this PR mergable. Would you rebase and amend the commits of yours to get them signed off (`git co --amend -s`)? That way DCO would be satisfied I think.

sam-19:
Okay, I tried to amend the commits by rebasing, which went to hell as usual. Now the rebase is stuck in an error state that I cannot resolve. I'm afraid this PR may be lost. I will pick the file changes into another PR sometime later when I have time.

christianlupus:
I just reverted your rebase as I had the original ones still at hand. If you are still mid-rebase, yopu can abort it via `git rebase --abort`. I will do some cleaning up and then help you with the rebase.

christianlupus:
I rebased and signed off on your behalf. I hope this was ok for you.

sam-19:
Thanks, that is completely ok. I don't know if it is a problem with VSCode's integration to git or what, but my rebases often get stuck in a loop when there is a merge conflict, and there seems to be no no way around it.

PR #327, opened 2020-10-04
Change RegEx for yield value to match digits in json-ld less restrictive

zwoabier:
According to https://schema.org/recipeYield the yield value is specified like "recipeYield": "1 loaf"
The suggested RegEx  /(\d*)/ is a more generic approach to find digits in the recipeYield string instead of  /^.*?(\d*)/ 

For example this works for chefkoch.de ("recipeYield": "6 Portion(en)") and fittastetic.com ("recipeYield": ["2","2 Portionen"]).
The old RegEx /^.*?(\d*)/ worked on chefkoch.de but not e.g. on fittastetic.com

Because we have the if function in the next line (221) that deals with several occurrences and picks the first one, it should be save to do it that way.

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -214,10 +214,11 @@ class RecipeService
 
         $json['recipeCategory'] = $this->cleanUpString($json['recipeCategory'], false, true);
 
+        
         // Make sure that "recipeYield" is an integer which is at least 1
         if (isset($json['recipeYield']) && $json['recipeYield']) {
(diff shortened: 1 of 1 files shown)


christianlupus:
@zwoabier Could you amend the commit and sign it off (see [check](https://github.com/nextcloud/cookbook/pull/327/checks?check_run_id=1205221924)).

Otherwise it looks good to me. I'd merge and see if any complaints are coming in, but I doubt it.

PR #347, opened 2020-10-20
Add workaround for recipeInstructions

victorjoos:
Some json recipes (like on https://www.colruyt.be/fr/en-cuisine/meli-melo-de-legumes-oublies-au-chevre) don't contain the array directly, but have it inside the 'itemListElement'. While this is clearly a mistaken understanding of the spec, this is a workaround that shouldn't impact existing json parsing.

Thank you very much for the project and the work !

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -295,6 +295,11 @@ class RecipeService
         // Make sure that "recipeInstructions" is an array of strings
         if (isset($json['recipeInstructions'])) {
             if (is_array($json['recipeInstructions'])) {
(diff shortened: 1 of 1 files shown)


christianlupus:
In general this looks good to me. However, there are multiple things to consider/extend here:

1. The page should be made aware to stick with the schema.org standard. 
2. The parsing is not ideal at this location. The standard allows to parse the instructions as a `ItemList` of `HowToStep`s. We are not capable to parse that at the moment.

During the rework towards point 2 this issue might come up again. Thus, I am merging here in a quick way.

PR #359, opened 2020-10-27
Feature/check recipe edits before leaving

### LINKED ISSUE #310: Show warning when navigating away from recipe editing window
## Current state

When accidentally navigating to a different site while in recipe-editing mode, all entered information is lost. 

This can happen easily, e.g.,  on a notebook, when scrolling (up/down) on a touch pad is falsely registered as a right swipe resulting in a history-back event or on a tablet when the screen space is limited and a tap on a Nextcloud app icon is registered contrary to the user's intent.

## Proposal

Probably the simplest solution would be to show a popup where one can select "Discard changes", "Save changes", or "Cancel" (for returning to edit mode).

An alternative solution could be an auto-save feature, when leaving the editing page. This would require a conscious click on a delete button to remove changes.

seyfeb:
Before leaving the edit page, the current edited recipe data is compared to the initial data. If there are changes, the user is prompted if she wants to leave the page.

Padding zeros in prep, cooking, and total time are ignored.

Closes #310

### DIFF
--- a/src/components/RecipeEdit.vue
+++ b/src/components/RecipeEdit.vue
@@ -80,6 +80,13 @@ export default {
         addEntry: function(field, index, content='') {
             this.recipe[field].splice(index, 0, content)
         },
+        beforeWindowUnload(e) {
+          if (this.confirmStayInEditedForm()) {
+            // Cancel the window unload event
+            e.preventDefault()
+            e.returnValue = ''
+          }   
+        },
         /**
          * Compares initial with current recipe. Returns true if they are the
          * same. Ignores padding that has been added to the hours and minutes
@@ -281,6 +288,9 @@ export default {
         })
         this.savingRecipe = false
     },
+    beforeDestroy() {
+      window.removeEventListener('beforeunload', this.beforeWindowUnload)
+    },
     // We can check if the user has browsed from the same recipe's view to this
     // edit and save some time by not reloading the recipe data, leading to a
     // more seamless experience.
@@ -331,6 +341,9 @@ export default {
         if (this.$window.shouldReloadContent(from.fullPath, to.fullPath)) {
             this.setup()
         }
+    },
+    created() {
+      window.addEventListener('beforeunload', this.beforeWindowUnload)
     },
 
 }
--- a/src/components/RecipeEdit.vue
+++ b/src/components/RecipeEdit.vue
@@ -80,6 +80,63 @@ export default {
         addEntry: function(field, index, content='') {
             this.recipe[field].splice(index, 0, content)
         },
+        /**
+         * Compares initial with current recipe. Returns true if they are the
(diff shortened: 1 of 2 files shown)


christianlupus:
I requested a review from @sam-19 as he is much more into the world of JS/Vue than me. From my (as a non-authority in this field) it looks good.

christianlupus:
@seyfeb I changed the CI a bit to require a small comment on the `CHANGELOG.md`. Currently this PR does not allow to recheck against the most current master thus the (currently failing) test is not shown. Please consider rebasing to the current master branch. If you have issues with that, I can do it if you like. Alternatively you could merge `master` to your current work. Thank you

seyfeb:
Rebased & updated changelog!

src/components/RecipeEdit.vue
```diff
@@ -201,20 +271,25 @@ export default {
     mounted () {
         // Store the initial recipe configuration for possible later use
         if (this.recipeInit === null) {
-            this.recipeInit = this.recipe
+            this.recipeInit = Object.assign({}, this.recipe)
```
seyfeb, line 274: @sam-19 I think this should circumvent the "passing by reference" you mentioned [here](https://github.com/nextcloud/cookbook/pull/378#issuecomment-724768627)

sam-19: Yes, I didn't notice there were two separate PRs. This will probably be enough.

seyfeb: In this PR I actually require the initial recipe state to check if the current recipe state has changes. The other PR is just a small fix to enable recipe creation at all until this PR is reviewed and accepted.

christianlupus:
I just rebased to the current state to have less issues during final merge and more up-to-date codebasee in general.

seyfeb:
@sam-19

I’m considering to take a different approach from this PR, i.e., not compare the initial recipe state with the current state (as this requires ugly deep comparison of js objects and arrays) but rather to set a "dirty" flag whenever the user makes changes.

While looking at this approach, I noticed that the editing components (`RecipeEdit`, `EditInputField`, ...) are tightly coupled by referencing the parent component in, e.g., the `EditInputField` component. I’d like to restructure it in a way that the dependency only goes one way: The child components should emit an event with an update when they have been changed by user input. That way the parent (`RecipeEdit`) can handle updates to, e.g., set aforementioned "dirty"-flag, padding time strings, etc. This would also allow more flexible ways of data binding and reusing the components in different places.

Would you agree that this is a sensible approach? 

btw, are you also in the matrix channel? If so, what’s your name there? :)

christianlupus:
@seyfeb I assume this PR is obsolete by now, right?

seyfeb:
That’s correct. I was just waiting for the other one to be closed!

Issue #332, opened 2020-10-09
Refactor URL endpoints


christianlupus:
Some of the endpoints are not well named or well structured. An example is given in [this issue discussion](https://github.com/nextcloud/cookbook/issues/163#issuecomment-602488329).
The endpoints should be reworked and used correctly throughout the app.

seyfeb:
As I see it, we have have as our main data models `Recipe`s and (considering #340) `Cookbook`s (in the future), and maybe `Category`s and `Keyword`s. What do you think about allowing the creation of categories and keywords without any recipes assigned?

I would then propose to have endpoints for (most) CRUD operations on these objects. There are other things one may like to use for filtering recipes such as cooking time, ingredients, tools, author, language, ... but they probably should be included as additional information and not become separate endpoints. 

[In the Nextcloud documentation on routing](https://docs.nextcloud.com/server/latest/developer_manual/basics/routing.html#registering-resources) there is an example for a resource which exposes endpoints to

- Get all entries
- Get one entry by id
- Create an entry
- Update an entry
- Delete an entry

This is currently used for the recipes, although, at the moment, the `MainController` contains some recipe-related stuff as well, which is handled by the `RecipeController`. I guess most of it is old and unused and requires some cleanup (also in `routes.php` unused routes should be removed).

At the same time I would propose to create separate Controllers for each of our resources to get a clearer structure in the code base.

christianlupus:
> What do you think about allowing the creation of categories and keywords without any recipes assigned?

Then we need to rework the database first (see #300).

> At the same time I would propose to create separate Controllers for each of our resources to get a clearer structure in the code base.

Definitively. This is part of the refactoring process of the codebase.
We should also consider moving to versioned API endpoints sooner than later (especially before any stable V1.0). Then we can have legacy controllers if the API endpoints need to change anytime in the future.

Issue #299, opened 2020-08-14
Refactoring of the services in the backend


christianlupus:
**Is your feature request related to a problem? Please describe.**
The idea behind a service is a single topic. The current services do not stick to this pattern. In fact, when #298 should be implemented, clearly separated and small classes are needed to make the test useful.

**Describe the solution you'd like**
A complete refactory of the services and rework of current PHP file structure would be involved. First, logical parts should be identified and these should then be implemented piece by piece.

I started to work on some of these refactorings due to my work in PR #297. However I did not want to blow up that PR too much and also not blow up any currently ongoing development.
A first step might be to define a rough time schedule to synchronize any currently ongoing development.

PR #362, opened 2020-10-29
Allow selecting recipes with unassigned category

### LINKED ISSUE #326: 'None' category selection
when i am importing several recipies, i sometimes loose track on which categories I have added the new recipies to.  it ends up being time consuming to track down those items that are not categorized yet.  is there a way to add a filter for uncategorized recipies?

seyfeb:
Added item for viewing uncategorized recipes in the navigation.

Closes #326

### DIFF
--- a/lib/Db/RecipeDb.php
+++ b/lib/Db/RecipeDb.php
@@ -188,11 +188,13 @@ class RecipeDb {
     
     /**
      * @throws \OCP\AppFramework\Db\DoesNotExistException if not found
+     *
+     * Using '_' as a placeholder for recipes w/o category
      */
     public function getRecipesByCategory(string $category, string $user_id) {
         $qb = $this->db->getQueryBuilder();
 
-        if ($category != 'None')
+        if ($category != '_')
         {
             $qb->select(['r.recipe_id', 'r.name'])
                 ->from(self::DB_TABLE_CATEGORIES, 'k')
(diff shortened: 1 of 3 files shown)


christianlupus:
I just had a quick look at the changes. In fact I have a bit of a bad feeling here. Apart from the fact that using an arbitrary symbol as a special value for no category, I found these lines: [[1](https://github.com/nextcloud/cookbook/blob/master/src/components/AppNavi.vue#L243-L245)], [[2](https://github.com/nextcloud/cookbook/blob/master/lib/Db/RecipeDb.php#L434-L437)]. These are using a star as symbol (neither really good).

I have to admit, that I have not yet tested the changes.

My suggestion is to stick with either symbol (makes no difference to me). Then in #332, #299 and #365 there is the chance to define a more stable API. Any comments/suggestions on that by anyone?

seyfeb:
I have seen the `*` being used as well, but had the feeling using this would be even worse. I would rather interpret it as a wildcard, matching every category. I didn’t want to break anything else though, so I just used `_` as a proposal. 

Obviously this would conflict with a category named “_”, so a new API endpoint might be helpful. I didn’t want to start on this as @christianlupus was discussing API changes somewhere else, anyway, if I remember correctly.

christianlupus:
You remember correctly, apart from #342 I was considering to do quite some backend refactory. If you do not mind that these changes are overwritten soon, I'd test them and merge until the more thorough rework is finished.

seyfeb:
No that’s fine. I just thought it might be nice to have this option even "sooner" ;) And to close the corresponding issue.

christianlupus:
OK, a quick test shows that it seems to work so far. Will merge that and hope for the best. :smile:

PR #373, opened 2020-11-05
Added tag/keyword cloud to recipe view

### LINKED ISSUE #247: Show keywords in recipe
I'd found it helpful to have all assigned keywords in the header of the recipe visible (and klickable).

Not only via "edit" accessible.

Maybe related to #188

seyfeb:
Added list of keywords below the title in the recipe view. Klicking keyword routes to recipe list with corresponding tag.

Closes #247

### DIFF
--- a/src/components/AppControls.vue
+++ b/src/components/AppControls.vue
@@ -13,7 +13,7 @@
             </Breadcrumb>
             <!-- SEARCH PAGE -->
             <Breadcrumb v-if="isSearch" class="not-link" :title="searchTitle" :disableDrop="true" />
-            <Breadcrumb v-if="isSearch && $route.params.value" class="active" :title="$route.params.value=='_'?'None':$route.params.value" :disableDrop="true" />
+            <Breadcrumb v-if="isSearch && $route.params.value" class="active" :title="$route.params.value=='_'?'None':decodeURIComponent($route.params.value)" :disableDrop="true" />
             <!-- RECIPE PAGES -->
             <!-- Edit recipe -->
             <Breadcrumb v-if="isEdit" class="not-link" :title="t('cookbook', 'Edit recipe')" :disableDrop="true" />
@@ -181,6 +181,8 @@ export default {
                 return t('cookbook', 'Recipe name')
             } else if (this.$route.name === 'search-tag') {
                 return t('cookbook', 'Tag')
+            } else if (this.$route.name === 'search-tags') {
+                return t('cookbook', 'Tags')
             } else {
                 return t('cookbook', 'Search for recipes')
             }
--- a/src/components/AppControls.vue
+++ b/src/components/AppControls.vue
@@ -179,8 +179,6 @@ export default {
                 return t('cookbook', 'Category')
             } else if (this.$route.name === 'search-name') {
                 return t('cookbook', 'Recipe name')
-            } else if (this.$route.name === 'search-tag') {
-                return t('cookbook', 'Tag')
             } else if (this.$route.name === 'search-tags') {
                 return t('cookbook', 'Tags')
             } else {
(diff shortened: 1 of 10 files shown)


christianlupus:
I just looked at your code. I found that the code uses the endpoint `/api/search/<tag>`. This is the general search query. So, if the tag was found e.g. in another recipe *name* that recipe will be returned as well.

Was this done by intention or just by chance?

If not done by intention, you might need to register a corresponding search functionality in [the search component](https://github.com/nextcloud/cookbook/blob/master/src/components/SearchResults.vue#L27-L29). The backend needs to be adopted as well (registering a route, adding a controller method and implementing the search itself).

seyfeb:
Good spot, I figured the `/tag/X` endpoint searches for the keyword. This will require a little bit more work then, so I need to have a look at this later. This PR is WIP then!

christianlupus:
I just opened a PR against your fork @seyfeb with the backend work done (at least until the point I assume it is working, testing needs to be done).

seyfeb:
I already did this in parallel, although I went a slightly different route by not using the /tag/ endpoint and instead introduced a new /tags/ endpoint which takes a comma-separated list of tags which must all exist in a recipe to be returned ("AND" operation). 

This allows for single tag searches but is more flexible.

christianlupus:
OK, I'll let you do your work until you are done.

BTW: [These lines](https://github.com/nextcloud/cookbook/pull/373/files#diff-924641363feab47bb47e36cb2cc1eeb9fb0b259f4d62d4288e9d58c4fbdaf3e9R253) might have an issue with the colon in the parameter's name.

seyfeb:
I’m glad about any help! ;) 
Should we support your endpoint, too? Then I would merge your PR.
Apart from that, I think this PR should be more or less complete then. Depending on your answer I would remove the WIP state.

christianlupus:
I already closed the PR. I think the API should be kept as clean as possible, aka drop my suggestion, please.

christianlupus:
I requested @sam-19 as requestor as I am not very well suited for vue.js currently. I will look at the code soon as well.

src/router/index.js
```diff
@@ -29,6 +29,7 @@ const routes = [
     { path: '/name/:value', name: 'search-name', component: Search, props: { query: 'name' } },
     { path: '/search/:value', name: 'search-general', component: Search, props: { query: 'general' } },
     { path: '/tag/:value', name: 'search-tag', component: Search, props: { query: 'tag' } },
```
christianlupus, line 31: This line might be obsolete. Of you define a new route `/tags/:value`, we can drop `/tag/:value` as I understand it.

src/components/AppControls.vue
```diff
@@ -181,6 +181,8 @@ export default {
                 return t('cookbook', 'Recipe name')
             } else if (this.$route.name === 'search-tag') {
                 return t('cookbook', 'Tag')
```
christianlupus, line 183: This should be obsolete as well.

christianlupus (approved):
The PHP part seems good. I have two comments regarding Vue.JS but I am no expert in thsi field.

src/router/index.js
seyfeb: that’s correct. I did not remove it initially because I wasn’t sure if we would want to leave this route in there. Considering the discussion in the PR, I’ll just remove it.

christianlupus:
Thank you for the contribution, @seyfeb. Now we can only wait for @sam-19.

sam-19 (approved):

sam-19:
No issues that I can see, good job! I have avoided using $emit to parent component, because I had some lingering hopes about porting my Vue apps using Vue Native until recently. But since that will probably never happen, using emits is "better" Vue programming and I'll start adopting them more myself.

Issue #188, opened 2020-04-06
List by Keywords / Category


boldt:
The old version used the "Keywords" to group the recipes. Because of this, I organized all my recipes by Keywords. The current version uses the "Category". Unfortunately, I cannot find a way to filter my recipes by a keywords any more. I really like the new category view, but there are use cases, where the keywords are still very useful:

* List all recipes with the keyword "barbeque", independent from their category. 
    * The "burgers" will be in the category "Main dish",
    * The "potato salad" will be in the Category "sides", and 
    * The "burger sauce" will be in the category sauces.
* List all recipes with the keyword "gluten free", independent from their category

I propose an additional filter to switch between "Keywords" and  "Category" view.

mrzapp:
I agree, we're missing the keywords as explicit UI. As I suggested before, @mcorteel should we provide a "tag cloud" somewhere in the interface? I think it would fit well underneath the title in the overview, like this:

![image](https://user-images.githubusercontent.com/2747428/78547862-0c4e1b80-7800-11ea-8133-82261bd94407.png)

What do you guys think?

hitam4450:
I liked the "old" keywords system much better compared the existing categories!
However, refining the current system would mot advisable !

boldt:
Under the title is a good idea. 

* **Case 1**: I click on the category "all recipies". The I click on the tag "barbeque". 
    * I'll see all barbeque recipies, independent from their category
* **Case 2**: I click on "main dishes". Then I click on the tag "gluten free"
    * I'll see all gluten free main dishes
* **Case 3**: I click on the category "all recipies". The I click on the tag "barbeque". Then on the tag "Gluten free"
    * I'll see all gluten free recipies for barbeque - I recomment an _AND_, not and OR_ to combine tags


then on "gluten free". 

The I click on "barbeque". I see all barbeque recipies

Will this be an

mrzapp:
@boldt case 3 just seems like an extended specification of case 2, unless I'm mistaken. Either way, I support using an AND operator, as well as keeping the category scope.

seyfeb:
This issue can be closed 🎊

PR #375, opened 2020-11-08
Show tag cloud in search results

seyfeb:
Regarding #188.

Added tag cloud between the breadcrumb nav and the recipe overview grid. What it currently does, is

- show all keywords of the displayed recipes
- keywords are selectable
- selecting one or multiple keywords shows only recipes in the current category/search that have all keywords attached

### Things to consider

- I changed the backend in such a way that not only the recipe name and id are returned anymore, but also a comma-separated string of keywords. The idea is to prevent querying the keywords of each recipe but only having a single database call.
- I wasn’t able to get a `GROUP_CONCAT` working with the QueryBuilder. This might be more efficient than my workaround, so if anyone knows how to get this working - feel free to contribute! 

Feedback, as always, welcome.

Screenshot 2020-11-08 at 14 30 50

### DIFF
--- a/src/components/AppIndex.vue
+++ b/src/components/AppIndex.vue
@@ -1,21 +1,33 @@
 <template>
-    <ul>
-        <li v-for="recipe in filteredRecipes" :key="recipe.recipe_id">
-            <router-link :to="'/recipe/'+recipe.recipe_id">
-                <img v-if="recipe.imageUrl" :src="recipe.imageUrl">
-                <span>{{ recipe.name }}</span>
-            </router-link>
-        </li>
-    </ul>
+    <div>
+        <ul v-if="keywords.length" class="keywords">
+            <RecipeKeyword v-for="(keyword,idx) in keywords" :key="'kw'+idx" :keyword="keyword" v-on:keyword-clicked="keywordClicked(keyword)" v-bind:class="{active : keywordFilter.includes(keyword)}" />
+        </ul>
+        <ul class="recipes">
+            <li v-for="recipe in filteredRecipes" :key="recipe.recipe_id" v-show="recipeVisible(recipe.keywords)">
+                <router-link :to="'/recipe/'+recipe.recipe_id">
+                    <img v-if="recipe.imageUrl" :src="recipe.imageUrl">
+                    <span>{{ recipe.name }}</span>
+                </router-link>
+            </li>
+        </ul>
+    </div>
 </template>
 
 <script>
+import RecipeKeyword from './RecipeKeyword'
+
 export default {
     name: 'Index',
+    components: {
+        RecipeKeyword,
+    },
     data () {
         return {
             filters: "",
             recipes: [],
+            keywords: [],
+            keywordFilter: [],
         }
     },
     computed: {
@@ -35,6 +47,17 @@ export default {
     },
     methods: {
         /**
+         * Callback for click on keyword
+         */
+        keywordClicked: function(keyword) {
+            const index = this.keywordFilter.indexOf(keyword);
+            if (index > -1) {
+                this.keywordFilter.splice(index, 1);
+            } else {
+                this.keywordFilter.push(keyword)
+            }
+        },
+        /**
          * Load all recipes from the database
          */
         loadAll: function () {
@@ -42,6 +65,7 @@ export default {
             var $this = this
             $.get(this.$window.baseUrl + '/api/recipes').done(function (recipes) {
                 $this.recipes = recipes
+                $this.setKeywords(recipes)
                 deferred.resolve()
                 // Always set page name last
                 $this.$store.dispatch('setPage', { page: 'index' })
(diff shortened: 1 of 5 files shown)


src/components/AppIndex.vue
```diff
@@ -65,35 +119,43 @@ export default {
 
 <style scoped>
 
-ul {
+ul.keywords {
+    display: flex;
+    flex-wrap: wrap;
```
christianlupus, line 124: This will make the comment swap around.

src/components/RecipeKeyword.vue
```diff
@@ -0,0 +1,49 @@
...
+
+.active li {
+    background-color: var(--color-primary);
+    color: var(--color-primary-text);
+}
+
+li:hover, .active li:hover {
+    border: 1px solid var(--color-primary);
+}
+</style>
```
christianlupus, line 49: This file might conflict with #373.

christianlupus (approved):
The backend code looks good to me. There is no way to use `GROUP_CONCAT` as we are not based on MySQL but a generic SQL driver.
Regarding the Vue.JS I cannot confirm everything as I am not an expert.

One thing I found during testing the functionality:
When one filters the search result by keyword further, it is possible to generate an empty result. This might be surprising for the user. We might want to hide all keywords that are no longer present in the currently filtered view. However, I suggest to keep the current behavior and open an issue/additional PR to solve that bug. Otherwise we might never finish this PR.

christianlupus:
I added a rewiew request for @sam-19 as he is our Vue.JS expert.

Thanks @seyfeb for the contribution.

src/components/RecipeKeyword.vue
seyfeb: yes and no. it should be identical 🤞

src/components/AppIndex.vue
```diff
@@ -52,6 +76,36 @@ export default {
...
+            if ((recipes.length) > 0) {
+                recipes.forEach(recipe => {
+                    if(recipe['keywords']) {                    
+                        recipe['keywords'].split(',').forEach(kw => {
+                            if(!this.keywords.includes(kw)) {
+                                this.keywords.push(kw)                           
+                            }
+                        })
+                    }                    
+                })
```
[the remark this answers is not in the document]
seyfeb: We might need to introduce a scroll view at some point, but for now this should be fine 

Screenshot 2020-11-08 at 22 35 08

src/components/AppIndex.vue
seyfeb: I’m not sure, which comment you mean? What do you mean by swap around?

seyfeb:
> One thing I found during testing the functionality:
> When one filters the search result by keyword further, it is possible to generate an empty result. This might be surprising for the user. We might want to hide all keywords that are no longer present in the currently filtered view. However, I suggest to keep the current behavior and open an issue/additional PR to solve that bug. Otherwise we might never finish this PR.

I ignored your suggestion and implemented your idea of disabling keywords which are not contained in visible recipes!

src/components/AppIndex.vue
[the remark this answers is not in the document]
christianlupus: That will depent how many recipes you have in your cookbook. Some pages set a certain amount of keywords to the tag per recipe. Of course, there will be similarities but for a few 100 recipes the amount of keywords might become significant. But you are right, this might be ok if done later.

src/components/AppIndex.vue
christianlupus: Sorry, typo. I meant **wrap**.

src/components/RecipeKeyword.vue
christianlupus: True, but for the merging we might have a close look for conflicts. No big deal if it's only whitespace or the like.

christianlupus:
Apart from my inability to review the Vue code in detail, it seems to work for me ;-)

src/components/AppIndex.vue
seyfeb: Yes, but that’s good, isn’t it? If the keywords don't fit in a single line, the user is still able to see them as in [this picture](https://github.com/nextcloud/cookbook/pull/375#discussion_r519480331).

sam-19:
I'm not sure why, but I am not able to test this PR out. When I try to search for recipes, I get the error
`Undefined property: OCA\Cookbook\Service\DbCacheService::$user_id at /var/www/html/custom_apps/cookbook/lib/Service/DbCacheService.php`
on the following lines:
https://github.com/seyfeb/cookbook/blob/481d7be78ec01fbacc0fceb4b0a2529f613fb7dc/lib/Service/DbCacheService.php#L347
https://github.com/seyfeb/cookbook/blob/481d7be78ec01fbacc0fceb4b0a2529f613fb7dc/lib/Service/DbCacheService.php#L355
https://github.com/seyfeb/cookbook/blob/481d7be78ec01fbacc0fceb4b0a2529f613fb7dc/lib/Service/DbCacheService.php#L383

This in turn triggers an uncaught type error
`Trying to access array offset on value of type bool at /var/www/html/custom_apps/cookbook/lib/Db/RecipeDb.php#478`

Am I the only one having this problem? Maybe it's something with my setup?

seyfeb:
I don’t think I can reproduce this. Should I see this in the server logs? I’m using @christianlupus docker setup. Where would I have to look? Also, I did not touch the `DbCacheService`.

sam-19:
Probably something weird going on with my dev setup. I can see how it works on the index page though and it looks cool :) There are some inconsistencies with line endings in the script sections, but if you wouldn't mind removing those few semicolons here and there, I would definitely give this a thumbs up!

seyfeb:
I removed the semicolons I could find from the updated vue components. Jumping back and forth between languages is annoying 😓

sam-19 (approved):

Issue #131, opened 2020-01-01
Support "dateCreated" field


spikusmax:
Hi there,

I would love to be able to record dates recipes were made.  With a view to future features such as searches based on dates, for the scenario: "Dang! What was that recipe I made for XYZ's birthday in 2015?" 

Cheers,

Spike

mrzapp:
This should be added to the RecipeService and the interface. @Teifun2 would you be able to look into this one?

Teifun2:
I like the idea of knowing when a recipe was added -> `dateCreated` but i don't think it totally solves the problem described in the issue. 

What if i don't know when i added the recipe to my collection but i know that i did it for my birthday? I cannot find the recipe with this time frame then. In theory we would need the possibility to add "creations" of each recipe, in a way that whenever i cooked the recipe i could say that i have done so.

This could also mean that we could allow images for each cooked recipe and maybe even comments? This could even replace the original idea of #8. 

With `workExample` we could add a new comment / image and date for each time the recipe was cooked. Most of the times comments would anyway be added after or during the cooking of a recipe.

As `workExample` is a type of [CreativeWork](https://schema.org/CreativeWork) which supports [Review](https://schema.org/Review) we could do it in a way that each "Cooking" :P is a review which can have a date and an image and comments and a rating, this would serve several purposes and somehow fits to the schema.org.

Edit: Just noticed that we dont need to go over `workExample` but cann add Reviews directly over `review` of a CreativeWork.

None the less i think it would be a good idea to add the `dateCreated` field and i certainly can do that at some point.

mrzapp:
@Teifun2 the "review" field is definitely interesting. Let's either keep the discussion on #8 or create a new issue if needed.

seyfeb:
I started working on this issue in PR #377

When creating own recipes it is obvious, what should happen. However, there are several edge cases when importing recipes from 3rd-party sites.

What should happen if an imported recipe

1. has no created & modified date
2. has no created but a correct modified date
3. has no created but an invalid modified date
4. has a created date but an invalid modified date
5. has a created date but a modified date which is before the created date
6. has an invalid created date (and different possibilities for modified date)
7. etc.

[...]

PR #377, opened 2020-11-09
Support dateCreated and dateModified fields

seyfeb:
Regarding #131, I added schema.org Recipe’s `dateCreated` and `dateModified` fields to the recipe view.

These also set/updated when creating or updating a recipe.

The recipe view 
- shows the created date if available
- shows the modified date if its
  - available
  - not the same as the dateCreated timestamp

Date & time are localized.

Screenshot 2020-11-09 at 14 38 48


christianlupus (commented):
The static code analysis looks good. I would like to try it in a test instance then I will approve it.

PR #379, opened 2020-11-09
Parse multiline ingredients and tools into separate fields

### LINKED ISSUE #325: Ability to paste multiple ingredients/instructions at once
**Use Case**
I have a recipe with ingredients with each ingredient on a separate line. I dare to say this is a common format that is used across many websites. Assume I have the recipe in a text file or I have found the recipe on a website that doesn't use schema.org. Now I want to enter it into the *Cookbook*. A similar case applies to the instructions, where steps can be separated by an empty line.

**Current Situation**
Right now I have to copy-paste each ingredient separately.

**Suggested Solution**
When multiple lines are pasted into the ingredient input field, they are automatically split by newline and entered as several ingredients. An example of such behaviour can be found in the *Gourmet Recipe Manager* where pasting multiline input into the ingredient field automatically adds multiple ingredients. A similar behaviour can be also found in *OpenEats* which uses a single form to enter ingredients one ingredient per line.

For instructions, it would be nice to split them into separate steps at empty lines.

seyfeb:
In the recipe editor, pasting content with newlines in a tools or ingredients field automatically creates new fields.

For textareas (recipe steps) this is not done.

Closes #325.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   [#366](https://github.com/nextcloud/cookbook/pull/366/) @christianlupus
 - Keyword cloud is displayed in recipe
   [#373](https://github.com/nextcloud/cookbook/pull/373/) @seyfeb
+- Pasted content with newlines creates new input fields automatically for tools and ingredients in recipe editor
+  [#379](https://github.com/nextcloud/cookbook/pull/379/) @seyfeb
 
 ### Changed
 - Switch of project ownership to neextcloud organization in GitHub
--- a/src/components/EditInputGroup.vue
+++ b/src/components/EditInputGroup.vue
@@ -4,8 +4,8 @@
         <ul ref="list">
             <li :class="fieldType" v-for="(entry,idx) in $parent.recipe[fieldName]" :key="fieldName+idx">
                 <div v-if="fieldName==='recipeInstructions'" class="step-number">{{ parseInt(idx) + 1 }}</div>
-                <input v-if="fieldType==='text'" type="text" @keyup="keyPressed" :value="$parent.recipe[fieldName][idx]" v-on:input="inputFieldChange" @paste="handlePaste" />
-                <textarea v-else-if="fieldType==='textarea'" :value="$parent.recipe[fieldName][idx]" v-on:input="inputFieldChange" @paste="handlePaste"></textarea>
+                <input v-if="fieldType==='text'" type="text" @keyup="keyPressed" :value="$parent.recipe[fieldName][idx]" v-on:input="handleFieldChange" @paste="handlePaste" />
+                <textarea v-else-if="fieldType==='textarea'" :value="$parent.recipe[fieldName][idx]" v-on:input="handleFieldChange" @paste="handlePaste"></textarea>
                 <div class="controls">
                     <button class="icon-arrow-up" @click="moveUp(idx)"></button>
                     <button class="icon-arrow-down" @click="moveDown(idx)"></button>
@@ -70,7 +70,7 @@ export default {
         /** 
          * Handle typing in input or field or textarea
          */
-        inputFieldChange: function(e) {  
+        handleFieldChange: function(e) {  
             // wait a tick to check if content was typed or pasted
             this.$nextTick(function() {
                 if (this.contentPasted) {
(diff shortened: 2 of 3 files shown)


seyfeb:
Now, this is done for textareas, too.

There are only frontend parts touched by this PR.  @sam-19, as the vue.js magician here, could you have a look at this?

CHANGELOG.md
```diff
@@ -48,6 +50,8 @@
   [#358](https://github.com/nextcloud/cookbook/pull/358) @christianlupus
 - Avoid recipe are no longer reachable when user changes locales
   [#371](https://github.com/nextcloud/cookbook/pull/371) @christianlupus
+- Creating new recipe not possible due to null reference
+  [#378](https://github.com/nextcloud/cookbook/pull/378/) @seyfeb
```
christianlupus, line 54: I think this is a leftover from another PR....

christianlupus (commented):
As I am no Vus.JS magician, I just looked over the general structure. I will request review by @sam-19.

CHANGELOG.md
seyfeb: Yeah, I merged #378 in there to be able to create new recipes at all. I can remove those lines.

christianlupus: So, #378 should be merged before. All right.

src/components/EditInputGroup.vue
```diff
@@ -19,42 +19,115 @@
...
+            let $ul = $li.parents('ul')
+
+            if (this.fieldType === 'textarea') {
+                for (let i = input_lines_array.length-1; i >= 0; --i)
+                {
+                    if (input_lines_array[i].trim() == '') {
+                        input_lines_array.splice(i, 1)
+                    }
+                }
+            }
```
sam-19, line 115: Should this check apply to ingredients and tools as well? Or is there some particular reason why empty lines should be preserved?

sam-19:
Apart from the small comment above, it looks good to me!

src/components/EditInputGroup.vue
christianlupus: I think the idea is to have empty lines for paragraphs in the instructions if needed.

christianlupus: Although this could be related to #381

seyfeb: Yeah, I don’t know. My reasoning was that currently we have the possibility/workaround to separate ingredient groups by using `##`. So if there is an empty line, the user could insert a caption. While grouping instructions might have a use case as well, there is currently no nice way to reflect this in the UI.

sam-19: I get the idea (and it's not a bad one). I just have a concern that when copy-pasting lists of ingredients from web pages there could be empty lines in between just because there are some imperceptible HTML elements embedded into the list and the the user would have to manually remove one or two or more empty lines for every ingredient/tool that was added. Maybe adding the odd caption between pasting the lists of ingredients or even afterwards would be the lesser evil, or what do you think?

seyfeb: You have a point there. What do you think could be those “imperceptible HTML elements”? If I have a list, I can clean the lines to check against an empty string.

sam-19: Well, a site that I frequently use wraps ingredients in <p></p> tags. When I copy-paste a list of ingredients from that site, I get an empty line between every ingredient. So I would have to manually remove all those empty lines whenever I used that site. Just removing the check for `fieldType==='textarea'` would automatically remove all those empty lines from my ingredient lists.

seyfeb: Yes, true, I was thinking of some invisible special characters. Empty lines, or essentially everything that is removed by a `trim()` is easy :)

src/components/EditInputGroup.vue
```diff
@@ -19,42 +19,115 @@
...
+                        }
                     }
-                    window.clearInterval(focusMonitor)
-                }
-                failSafe -= 100
-                if (!failSafe) {
-                    window.clearInterval(focusMonitor)
-                }
-            }, 100)
+                })
```
seyfeb, line 61: @sam-19 btw, I'm not sure if the "dirty hack" part for focussing the new field was yours, but I just wanted to point this section out for deeper review and maybe future reference. for me this seems to work fine! :)

sam-19: That was probably just grabbed from some old project when I wasn't yet familiar with the tick-based update method Vue uses. $nextTick is definitely the way to go here.

src/components/EditInputGroup.vue
```diff
@@ -19,42 +19,114 @@
...
+            // get data from clipboard to keep newline characters, which are stripped
+            // from the data pasted in the input field (e.target.value)
+            var clipboardData = e.clipboardData || window.clipboardData;
+            var pastedData = clipboardData.getData('Text');
+
+            let input_lines_array = pastedData.split(/\r\n|\r|\n/g)
+            if ( input_lines_array.length == 1) {
+                return
+            }
+            e.preventDefault();
```
sam-19, line 102: A couple of semicolons again, other than that it looks good to me! :)

PR #381, opened 2020-11-10
Markdown for description

TheMBeat:


sam-19:
Vue Showdown looks promising at a glance. NC Text (being the main official NC app that utilizes markdown, I imagine) uses markdown-it, but my impression is that they have not completely settled on a rendering library yet. So treading carefully and enabling only basic markdown syntax would probably be the best way to go for now, to avoid causing confusion with different sets of features in different apps.

TheMBeat:
What do you mean by enabling only the basic Markdown syntax? We can talk about the config in the main.js.
Currently there is no markdown editor available in the recipe editor, because the styling did not fit to the other input fields. But in the PR is one that is not activated.

sam-19:
I mean that the 'github' flavor set for the editor supports at least two features that I believe NC Text does not: tables and emojis. Enabling features that are not available in other NC apps could cause confusion when, for example, copying and pasting content from one app to another. Just saying we should be mindful of this.

TheMBeat:
That makes sense. I will work on the settings again.

@sam-19: A bit offtopic, but could you take a look at my repo in the NewRecipeScraper branch to see if it would be possible to add this?

seyfeb:
I agree that we should clearly define the set of Markdown commands. Maybe we can consider using [MultiMarkdown](https://en.wikipedia.org/wiki/MultiMarkdown), which is also rather widespread?

sam-19:
> 
> 
> @sam-19: A bit offtopic, but could you take a look at my repo in the NewRecipeScraper branch to see if it would be possible to add this?

Wow, that's quite a bit to look at! There are some things that I'm not familiar with, but using a javascript parser for recipes is not something I would have a problem with (and I wouldn't mind being able to import recipes from my favorite recipe site allrecipes.org!). @christianlupus is really the guy to talk to though, since his more familiar with the current parser. Maybe you should open an issue or PR for discussion, so this one won't get derailed?

TheMBeat:
> > 
> 
> > 
> 
> > @sam-19: A bit offtopic, but could you take a look at my repo in the NewRecipeScraper branch to see if it would be possible to add this?
> 
> 
> 
> Wow, that's quite a bit to look at! There are some things that I'm not familiar with, but using a javascript parser for recipes is not something I would have a problem with (and I wouldn't mind being able to import recipes from my favorite recipe site allrecipes.org!). @christianlupus is really the guy to talk to though, since he's more familiar with the current parser. Maybe you should open an issue or PR for discussion, so this one won't get derailed?

The problem is that it doesn't work properly because there are CORS problems with the fetch.  I was hoping you could help.

TheMBeat:
With Vue-Showdown most [options](https://github.com/showdownjs/showdown#valid-options) like table or emoji are disabled by default. For Style we could use Vanilla or Original instead of GitHub. With these settings we should be close to the Nextcloud Text app or am I wrong?

christianlupus:
I would not worry too much about the set of Markdown commands available. It is merely a question of documentation. No user will open an issue stating that a certain feature has to be removed.
The bigger question is to ask which flavor to use to be maximally compatible with other (NC) apps. Just my 50ct.

I requested formal revieew of @sam-19 to check for correct integration. The question of the flavor might of course be an open discussion, as I think.

src/main.js
```diff
@@ -171,6 +177,14 @@ import AppMain from './components/AppMain'
     Vue.prototype.$window = window
     Vue.prototype.OC = OC
 
+    //Markdown for Vue
+    Vue.use(VueShowdown, {
+        //set default flavor for Markdown
+        flavor: 'github',
```
seyfeb, line 183: If we see the use case that our MD content is used in the other apps, I would suggest to use the `vanilla` flavor. [It seems]() that the NC Notes devs are going to support only a very basic set of Markdown features. 

If it’s only the other way round (copying from NC Notes to Cookbook), I would vote for a more feature-rich flavor.

src/components/EditInputField.vue
```diff
@@ -3,8 +3,9 @@
         <label>
             {{ fieldLabel }}
         </label>
-        <input v-if="fieldType!=='textarea'" :type="fieldType" v-model="content" @input="handleInput" />
-        <textarea v-if="fieldType==='textarea'" v-model="content" @input="handleInput" />
+        <markdown-editor class='editor' v-if="fieldType==='markdown'" :type="textarea" v-model="content" @input="handleInput" toolbar='' />
```
seyfeb, line 6: a) This is not set for the description field in the RecipeEdit component. 

b) Also it would be nice to activate the toolbar with some basic functionality (e.g., `bold italic | numlist bullist quote | preview`) to allow users who are not familiar with the Markdown syntax to easily add some formatting to their description.

src/main.js
seyfeb: Independently of other NC Apps: The supported flavor of the editor and viewer component should probably be the same

TheMBeat:
Hey @seyfeb my time is very limited. Also, I'm not very strong in CSS and had tried to use the toolbar on the first commit and failed, so it's disabled. 
I like your idea with the custom component based on Vue Showdown, unfortunately I won't be able to do it anytime soon.

PR #402, opened 2020-11-19
Added multiselection for categories and keywords to recipe editor

### LINKED ISSUE #313: Improved category selection for recipes
## Current state

In recipe-editing mode the complete category needs to be typed in by hand. Typos result in newly created categories requiring manual fixing of recipes.

## Proposed state

When typing in the categories field the typed string is compared to all existing categories. Initially the comparison could be done character by character, later maybe with some fuzzy search. The detected category or a list of best results should be presented to the user who can select the correct one by clicking with the mouse or selecting with the arrow keys + return.

## Alternative

An alternative for the selection would be a dropdown list in which all existing categories are presented. The user can select from the list to fill the category field of the recipe.

## Additional Considerations

This suggestion could also be relevant for selecting ingredients, keywords, and tools. Maybe a generic approach which is reusable for those cases might be a good thing.

seyfeb:
This PR adds a NC-styled `Multiselect` element for keywords and categories for faster insertion.

There is some collision with PR #386 but only for the keyword and category selection, which should be replaced by this one if accepted.

Closes #313 although for the future we might want to consider adding something similar for selecting ingredients, keywords, and tools as mentioned in the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
   [#390](https://github.com/nextcloud/cookbook/pull/390) @christianlupus
 - Automatic deployment of new releases to the nextcloud app store
   [#433](https://github.com/nextcloud/cookbook/pull/433) @christianlupus
+- Category and keyword selection from list of existing ones in recipe editor
+  [#402](https://github.com/nextcloud/cookbook/pull/402/) @seyfeb
 
 ### Changed
 - Switch of project ownership to neextcloud organization in GitHub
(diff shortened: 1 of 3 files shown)


seyfeb:
@sam-19 Could you have a look at this? This PR touches only frontend parts.

christianlupus (approved):
From my perspective, this looks good. I just checked in a dev environment.

sam-19:
This is a good change. I don't have time to properly test it in a dev setting, but don't see anything apparently wrong with the changes. See my comment about using an implicit value prop in #386.

PR #403, opened 2020-11-20
Optimized Webpack-Config for reduced bundle size

TheMBeat:


christianlupus:
As this is highly related to the JS/frontend code, I assume, @sam-19 is the best bet to get a grip on these changes.

Issue #300, opened 2020-08-14
Refactoring of the DB schema


christianlupus:
Currently the DB schema seems non-optimal solved.

**Describe the solution you'd like**

1. The keywords are associated with both a recipe and a user. A recipe would be sufficient.
1. The category is (by the schem.org standard) only one entry per recipe. This should be in a nullable column in the recipe table.
1. The recipes are indexed by the id of their parent folder and the user id. As the parent folder is already unique, the user id is not needed and should be dropped. This also makes the update faster as shared recipes are automatically updated for all users. 
1. Adding of indices if possible.
1. Avoid storing of series of strings (keyword as column), better use separate table with indexed foreign key + joins
1. Optionally enhance the schema to store more details for full-text search (see discussion on matrix channel from Aug 7th to 11th)

Potentially there are more optimizations to be carried out that come up during development.

PR #407, opened 2020-11-22
WIP: Create an abstraction layer for the database

christianlupus:
To allow for simpler development and future implemntation, this PR provides an internal API to the database.

It is related to #301.

Currently this PR is under development and should be considered Work in Progress. Suggestions and enhancements are welcome.

- [ ] creating valid architecture
- [ ] implemented features
- [ ] test code
- [ ] document interface
- [ ] write unit test files

### DIFF
--- a/lib/Db/CategoryDbWrapper.php
+++ b/lib/Db/CategoryDbWrapper.php
@@ -1,8 +1,7 @@
 <?php
 
-namespace  OCAA\Cookbook\Db;
+namespace OCA\Cookbook\Db;
 
-use OCA\Cookbook\Db\AbstractDbWrapper;
 use OCP\IDBConnection;
 use OCA\Cookbook\Entity\CategoryEntity;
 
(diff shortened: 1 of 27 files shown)


lib/Db/CategoryDbWrapper.php
```diff
@@ -0,0 +1,102 @@
...
+		
+		return $ret;
+	}
+	
+	/**
+	 * Store a single entity back to the database
+	 * @param CategoryEntity $category The entity to store
+	 */
+	public function store(CategoryEntity $category): void {
+		// We do not need to store anything here. The categories are just virtually generated.
```
seyfeb, line 61: What about creating categories as actual objects that can have, e.g., a description or other metadata? 😇

christianlupus: Simply put: No. :grin:

The current DB schema has only a category name stored in a mapping type table. I kept with that for now. See also #300.

I was planning to proceed as follows:
1. Create the entities and a DB abstraction layer parallel to the current main work.
2. Deprecate major parts of the current RecipeDbService and refactor the DbCacheService that use direct DB operations.
3. Migrate and refactor any current usage of any of these methods in the codebase
4. Migrate any pending PRs to the new system
5. Remove the direct DB access

After the DB access has been encapsulated as described here, we can add further setter/getter and fields in the DB without affecting the functionality of other parts.

PR #408, opened 2020-11-22
Improved Keyword filtering

seyfeb:
I made some changes to improve the keyword filtering in the recipe-list view. With this PR

- keywords show number of recipes that have the keyword assigned
- keywords are ordered by the number
- when keywords are selected, non-selectable keywords are moved to the end of the list
- selected keyword are moved to the beginning of the list for easy discovery
- a maximum size for the keyword-selection area is defined to prevent the page getting cluttered with many keywords. The area can be scrolled vertically to allow access to all keywords.

@sam-19 maybe you can review this PR?

### DIFF
--- a/lib/Db/RecipeDb.php
+++ b/lib/Db/RecipeDb.php
@@ -310,7 +310,7 @@ class RecipeDb {
 		$qb->setParameters($params, $types);
 		$qb->setParameter('user', $user_id, TYPE::STRING);
 
-		$qb->groupBy(['r.name', 'r.recipe_id']);
+		$qb->groupBy(['r.name', 'r.recipe_id', 'k.name']);
 		$qb->orderBy('r.name');
 
 		$cursor = $qb->execute();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,8 @@
   [#435](https://github.com/nextcloud/cookbook/pull/435) @christianlupus
 - Images in recipe list are lazily loaded
   [#413](https://github.com/nextcloud/cookbook/pull/413/) @seyfeb
+- Improved keyword filtering in recipe lists
+  [#408](https://github.com/nextcloud/cookbook/pull/408/) @seyfeb
 
 ### Fixed
 - Add a min PHP restriction in the metadata
--- a/src/components/AppIndex.vue
+++ b/src/components/AppIndex.vue
@@ -1,8 +1,16 @@
 <template>
     <div>
-        <ul v-if="keywords.length" class="keywords">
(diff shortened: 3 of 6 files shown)


christianlupus:
As this is mainly Vue-related stuff,  I added sam as a reviewer. I will have a look once I find a few spare minutes.

seyfeb:
I just converted this PR to a draft as I found, that the recipe search is not showing the correct keywords. I guess this is related to the php part, I’ll have a look at this later.

lib/Db/RecipeDb.php
```diff
@@ -310,7 +310,7 @@ public function findRecipes(array $keywords, string $user_id) {
 		$qb->setParameters($params, $types);
 		$qb->setParameter('user', $user_id, TYPE::STRING);
 
-		$qb->groupBy(['r.name', 'r.recipe_id']);
+		$qb->groupBy(['r.name', 'r.recipe_id', 'k.name']);
```
christianlupus, line 313: Why are you grouping by the keyword? A single keyword should be only uniquely in the DB associated with a recipe. So I do not get the intention behind this change.

christianlupus (approved):
Seems good to me. I checked the visual representation and it seems to work well.

christianlupus:
@sam-19 please give your statement. If you do not find the time, we can merge and fix bugs later if they happened...

lib/Db/RecipeDb.php
seyfeb: Otherwise each recipe is only returned with a single keyword.

Compare, e.g.,

```
SELECT `r`.`recipe_id`, `r`.`name`, `k`.`name` as `keywords` 
FROM `oc_cookbook_names` `r` 
LEFT JOIN `oc_cookbook_keywords` `k` ON k.recipe_id = r.recipe_id 
LEFT JOIN `oc_cookbook_categories` `c` ON r.recipe_id = c.recipe_id 
WHERE (r.user_id = "admin") 
GROUP BY `r`.`name`, `r`.`recipe_id`, `k`.`name` ORDER BY `r`.`name` ASC;
```

with

```
SELECT `r`.`recipe_id`, `r`.`name`, `k`.`name` as `keywords` 
FROM `oc_cookbook_names` `r` 
LEFT JOIN `oc_cookbook_keywords` `k` ON k.recipe_id = r.recipe_id 
LEFT JOIN `oc_cookbook_categories` `c` ON r.recipe_id = c.recipe_id 
WHERE (r.user_id = "admin") 
GROUP BY `r`.`name`, `r`.`recipe_id` ORDER BY `r`.`name` ASC;
```

seyfeb: You should have recipes with multiple keywords in your database to see the difference

christianlupus: Yes, you are completely right.

christianlupus:
This PR has conflicts with the lazy-loader feature recently added. @seyfeb sorry, you might need to update this as well.

Issue #406, opened 2020-11-22
Bigger Thumbnails


Teifun2:
**Is your feature request related to a problem? Please describe.**
Currently the image of a recipe is once scaled town and cut to a square to create a thumbnail. This Thumbnail has a very low resolution.

**Describe the solution you'd like**
For Apps or other Clients it would be great to have a wider variety of thumbnail Resolution, to suit different devices.

**Describe alternatives you've considered**
If we don't store higher resolution thumbnails we always need to download the full image for devices with a big resolution. This is traffic intensive but would work.

Related Issue on Android App:
https://github.com/Teifun2/nextcloud-cookbook-flutter/issues/28

If it would be ok to do that, i could create an initial pull request to implement that.

christianlupus:
If you want to take a first attempt, you are welcome to do so. There might be some changes soon (e.g. #340) but that will take some time until implemented. Just be aware to not put too much effort into it as it might get updated anyways.
Just assign yourself with this issue to mark it "in progress".

Teifun2:
Hmm the #340 Is very interesting.

It depends a little. Do you know how long it will take to incorporate these changes? If it is 1-2 Months i might wait for my changes but its something more like a year i probably will change it for the short term!

seyfeb:
@Teifun2 In #413, I implemented lazy-loading for thumbnails in the recipe lists. While already dealing with this stuff, I changed the default thumb size to 256x256. What resolution do you have in mind?

Teifun2:
@seyfeb Oh wow thats great to hear! I think bigger thumbnails are always "better" ^^
I was thinking that we could maybe add different "sizes" and load on demand what fits best. But thats quite a lot of work!
I'm very happy if i have 256 images :D

christianlupus:
We might want to define a set of sizes we want to have and create these in the background upon the implementation of #340.

PR #413, opened 2020-11-24
Lazy loading images in recipe list

### LINKED ISSUE #409: Slow loading Pages
Slow loading Pages.

For larger recipe collections, lazy-loading (Blure) the images for the recipe overview could improve the performance.

seyfeb:
This PR provides lazy loading of the images as discussed in #409 (currently only in the AppIndex) and also introduces a higher-resolved thumbnail as discussed in #406 (to be discussed).

The thumbnails are only loaded when they enter the viewport using the Intersection Observer API. The images are rendered/loaded in the following way

- a loading spinner is shown, while a small preview image (~1KB) is loaded from the server
- once this was loaded, the loading indicator is hidden and a blurry version of the image is displayed, while the full image is loaded in the background
- once the full image is loaded, the blurry version is replaced by the final image

This proposal needs to be tested and checked. 

Closes #409

CC: @sam-19


lib/Service/RecipeService.php
```diff
@@ -748,7 +752,7 @@ public function addRecipe($json) {
 			$thumb_image = new Image();
 			$thumb_image->loadFromData($full_image_data);
 			$thumb_image->fixOrientation();
-			$thumb_image->resize(128);
+			$thumb_image->resize(256);
```
christianlupus, line 755: This will default the thumb image size to 256x256. We should make sure, this is no problem as the old files are still 128x128.

christianlupus (commented):
Looks good in general. I have not yet tested it but I assume it is working.

lib/Service/RecipeService.php
seyfeb: I would like to put this out for discussion. Actually this change doesn’t make too much sense, since currently the images are shown in the lists in 105px by 105px - so 128 is more sensible in the sense that is less data to be transfered. However, there seems to be the need for larger thumbnails (#406). We could also keep the 128px thumbnail `thumb.jpg` and create a 256x256 version as `thumb@2x.jpg`

seyfeb:
> Looks good in general. I have not yet tested it but I assume it is working.

Before this gets merged, I’d prefer to have someone test it :)

lib/Service/RecipeService.php
christianlupus: I do not consider this too much of a problem to use 256x256. It is only a few kB per recipe.

christianlupus:
I just checked with an existing (dev) version. I found one issue yet open: The pre-existing images are not rerendered for the thumb16. So while the main images are loading, all recipes that were imported before the merge are showing a fork and knife in blurred version. 

I would consider this OK for now as it should not cause trouble as it is shown only a very limited amount of time. We should however prepare a migration step in #340 to make everything consistent.

seyfeb:
@christianlupus  I added a check if thumb16 is present and recreate the thumbnails if this is not the case.

seyfeb:
I also added the lazy-loading to the search results page.

sam-19:
I'm not familiar with the Lozad library, but the general idea seems good to me. I just wonder if the recipe links (in index and search results) should have been converted into a component of its own. This could possibly save some trouble and iteration, especially if recipe lists are are used in more places in the future. Then again, every new component increases complexity a bit, so it's probably not worth it yet. There is one import line ending in a semi-colon in LazyPicture.vue, though :)

seyfeb:
> I just wonder if the recipe links (in index and search results) should have been converted into a component of its own. 

Do you mean the complete tiles? I.e., Image plus title? I also thought about this and I guess that is a good idea. Especially if we want to change the layout at some point (e.g. square tiles with the image filling the tile and the title as an overlay) or even simply adding more information (category, number of stars, sharing status, cookbook name (if there are multiple at some point),  etc.) to it.

However, I don’t think this is the right PR to do this.

> There is one import line ending in a semi-colon in LazyPicture.vue, though :)

I’ll remove this!

christianlupus:
I was too quick with the merging, unfortunately. I had to revert and emerge to get the checks done right. Correct merge commit is 66f8023b.

Issue #414, opened 2020-11-26
Replace icons by SVG(s)


christianlupus:
The image at https://github.com/nextcloud/cookbook/blob/master/img/recipe-thumb.jpg (and possibly others) might be better off as SVG files that can be optimized and arbitrarily scaled.

TheMBeat:
Can't we just use app.svg on a grey background?

christianlupus:
Maybe with another foreground color, yes. But 500 Bytes are less than a few kB in `recipe-fill.jpg` or `recipe-thumb.jpg`.

TheMBeat:
@christianlupus
 What should the new SVGs look like? What background color and what color should the icon have?

christianlupus:
I have not extracted the RGB values but the SVG should be usable as a replacement for e.g. https://github.com/nextcloud/cookbook/blob/master/img/recipe-full.jpg. So should be straight forward.

TheMBeat:
Okay, i can do this next week.

PR #418, opened 2020-11-30
New thump and full image

TheMBeat:
Fix  [#415](https://github.com/nextcloud/cookbook/issues/414)


christianlupus (commented):
Why are two different SVG files required? The idea was to have a single well-scalable image.

Additionally, it might be a good idea to popp things up with CSS+Vue changes so that the existing images are replaced by an SVG image + gray background of appropriate size.

TheMBeat:
One SVG makes more Sense. I hadn't thought through to the end.  My mistake.  I'll improve it.

christianlupus:
I will have a look later as I am quite busy currently due to day-time job.

src/components/RecipeImages.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div v-if="$store.state.recipe.image" :class="{ 'collapsed': collapsed, 'printable': $store.state.recipe.printImage }">
+    <div v-if="$store.state.recipe.image" :class="{ 'collapsed': collapsed , 'printable': $store.state.recipe.printImage }">
```
christianlupus, line 2: This is a white-space only change. Is this really needed?

src/components/RecipeImages.vue
```diff
@@ -24,17 +24,20 @@ export default {
...
-	    text-align: center;
-	    margin-bottom: 1rem;
-	}
+    div {
+        display: block;
+        width: 100%;
+        text-align: center;
+        margin-bottom: 1rem;
+
+    }
```
christianlupus, line 33: What are these changes? I do not get the difference here.

src/components/RecipeImages.vue
```diff
@@ -24,17 +24,20 @@ export default {
...
     img {
         cursor: pointer;
         width: 100%;
         max-width: 950px;
+        background-color: #bebdbd;
     }
+
```
christianlupus, line 40: White-space only can be dropped.

christianlupus (changes requested):
There is a bug in the PR. When opening the cookbook app, I get these errors in the console:
Screenshot_20201206_170045

TheMBeat:
I look at this tomorrow.

TheMBeat (commented):
@christianlupus are the changes ok for you now?

christianlupus (approved):
Looks good and I see no issues here.

Teifun2:
This leads to a lot of problems in my App development.

The browser is totally happy if instead of a PNG he receives a SVG. But for normal App Components this always results in crashes deep in the library. As it gets a proper response but cannot parse the response into an image of any type.


Would it be ok if i changed the code so that the RecipeController returns 404 if no image is found. And the frontend itself adds the SVG in these cases? This seems a much cleaner "API" solution.

christianlupus:
@Teifun2 if this needs more discussion, we might be better of in a new issue instead of this PR.

Apart from that, can't you check the `Content-Type` header before?

PR #433, opened 2020-12-11
Create basic structure to allow automatic deployments

### LINKED ISSUE #233: Automatic deployment
We should set up an automatic deployment task whenever code is pushed to the `stable` branch

christianlupus:
This PR should allow deployments to the NC app store using simple pushes to the `stable` branch.

Closes #233


christianlupus:
@mrzapp I tried to use the REST API of the app store of NC. According to their policy, only maintainers and co-maintainers can upload new releases (in addition to the signature process using `openssl`). Currently, you are the only maintainer. All other people will get a `403 Forbidden` answer upon release upload.
I see 2 options here:
1. You provide a login token to your account of the NC app store as a secret in the repo (it is in your account settings under security)
2. You add another co-maintainer to the app who's token can be used.

I am not able to proceed here any further with the main task but need to know what way we should go.

mrzapp:
@christianlupus thanks for looking into this.

I was unable to find any way to add other accounts as co-maintainers, but I've provided an API token by the name "APP_STORE_TOKEN" in the repo secrets now

christianlupus:
As I cannot test the syntax without actually doing the release, I will have to hope for the best and give the PR a try. If it does not work, I will have to fix the issues starting from there.

Issue #284, opened 2020-07-03
Encoding errors with recipes from hellofresh.ch


clesmian:
**Description**
Importing recipes from hellofresh.ch works just fine, but there are encoding issues with special characters like äüöîé etc.

**Reproduction**
1. Import recipe from hellofresh.ch containing special characters (for example: https://www.hellofresh.ch/recipes/herzhafte-gnocchi-pfanne-5e136097a18baa02e540e58b) 
2. Check for example the headline of recipe 

**Expected behavior**
It should say "Käse-Mix", but it's displaying "KÃ¤se-Mix"

**Technical details**
As far as I can see the json included by hellofresh.ch is completely valid and everything works just fine, when you copy it from the pages source and put it into a json file inside of the recipe folder. 
I suspect that  somewhere along the lines of parsing the recipe there is an issue with incorrect encoding. 

Unfortunately I've not enough experience with php to debug the code myself and find the root cause of the issue.

glogiotatidis:
I get encoding errors from https://akispetretzikis.com/el/categories/soypes-2/karotosoypa which is pretty popular in greece. I believe the problem is that neither of those two websites declare document charset with `<meta charset="utf-8" />` and I guess php defaults to `latin` 

My suggestion is that cookbook defaults to utf-8 if no charset is defined.

burkhre2:
I get encoding errors for german 'umlaut' (e.g. instead of 'ü' I get 'Ã¼' )  from some webpages, e.g. https://migusto.migros.ch/de/rezepte/kuerbis-schoggi-kuchen where `<meta charset="utf-8" />` is specified in the page source. So the problem might be something else than just the default charset setting.

But another German page also with 'umlaut' characters e.g. https://cloud.marebu.ch/apps/cookbook/#/recipe/17082 is imported correctly.

cholletk:
Same there in french for fooby.ch (https://fooby.ch/fr/recettes/12340/roastbeef-aux-topinambours-et-au-sarrasin) and like @burkhre2 , there is <meta charset="utf-8"/> in soruce

christianlupus:
The problem with some of the named sites is that they do not contain valid HTML. Especially the definition of the content encoding is coming too late (must be within the first 1024 characters).

See:
[...]

PR #456, opened 2020-12-21
Fix encoding issue when source is encoded in utf8 (issue #284)

cholletk:
Adding utf8 decoding prior of editing recipe.

(and my IDE remove lot of spaces on empty lines, sorry for that)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,8 @@
   [#437](https://github.com/nextcloud/cookbook/pull/437) @christianlupus
 - Corrected bugs in CI system
   [#447](https://github.com/nextcloud/cookbook/pull/447) @christianlupus
+- Correct encoding error when importing UTF8 Encoded recipe
+  [#284](https://github.com/nextcloud/cookbook/issues/284) @cholletk
 
 ### Removed
 - Travis build system
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -426,7 +426,7 @@ class RecipeService {
 		// Make sure we don't have any encoded entities in the HTML string
 		$html = html_entity_decode($html);
 
-		// Convert utf8 entity 
+		// Convert utf8 entity
 		$html = utf8_decode($html);
 
 		// Start document parser
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -53,7 +53,7 @@ class RecipeService {
 
 		return $this->parseRecipeFile($file);
 	}
-	
+
 	/**
 	 * Get a recipe's modification time by its folder id.
 	 *
@@ -114,7 +114,7 @@ class RecipeService {
 		if (!$json) {
 			throw new Exception('Recipe array was null');
 		}
-		
+
 		if (empty($json['name'])) {
 			throw new Exception('Field "name" is required');
 		}
@@ -146,7 +146,7 @@ class RecipeService {
 						if (empty($img)) {
 							continue;
 						}
-		
+
 						$image_matches = [];
 
 						preg_match_all('!\d+!', $img, $image_matches);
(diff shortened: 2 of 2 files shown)


christianlupus:
OK, I just looked at your PR.

First, you committed quite many whitespace changes all over the files. This is not good as it causes merge conflicts with other changes in the future.
In general, this is no big deal, I will rebase/rework these changes before emerging into `master`.

More interesting is the fact that you unconditionally parse the HTML input through `utf8_decode`. I see the different cases here:

1. The page uses UTF8 and claims ISO-8859-1 or similar.
2. The page uses UTF8 and claims so.
3. The page uses ISO-8859-1 and claims so.
4. Any other encoding like ISO-8859-6 is either used or claimed.

Your fix obviously works for case 1. However, this is only a workaround for a broken web page.
For cases 2+3 the [manual](https://www.php.net/manual/de/function.utf8-decode.php) is not 100% clear but the user comments point towards the fact that the string gets corrupted. I assume double decoding of a string *will* destroy it.
If any encoding other than ISO-8859-1 is in the play, you would have to use `iconv()` to translate between arbitrary charsets.

@cholletk. am I missing here a point or your PR and you have more in-depth knowledge so you can explain what the idea was?

cholletk:
I agree with you on different uses cases. I think the best way to fix that in a cleaner way would be to use the declared charset in headers in order to perform conversion with iconv. 

One thing that I don't understand right now is why the content has to be converted to ISO-8859-1 (and how this charset is defined in order to do something adaptable). Do you have any idea on that?

christianlupus:
> One thing that I don't understand right now is why the content has to be converted to ISO-8859-1 (and how this charset is defined in order to do something adaptable). Do you have any idea on that?

I'd say this is not a requirement to go for ISO-8859-1. Most probably, the DOM parser does the charset conversion according to the internal structure. If that would not be true, I assume there would be many more complaints about wrong characters by now. In a small test using PHP, I parsed an HTML file with different charsets. It seems that ISO-8859-1 (or something similar to it) is a fallback if no valid charset was found within the first 1024 bytes. If the charset was found until then the conversion is done by the `DOMDocument` decoder. If not found until the limit is reached the fallback (whatever it is) takes effect causing the weird letters to apply.

Internally, PHP used normally single-byte characters. If dealing with Unicode strings, you need to use the `multibyte_*` functions and be very careful about what you are doing. I *guess* that at least part of the parser is written in PHP and thus uses single-byte chars as default.

We could use [this trick](https://www.php.net/manual/de/domdocument.loadhtml.php#95251) to make the parsing aware of the correct charset...

damda58:
image
not fix for me 
recipe on this link : https://www.hellofresh.fr/recipes/burger-de-boeuf-oignons-confits-lard-comte-60882ee1a3d000224e2703f2/

christianlupus:
> Would it not be possible to let the user choose whether or not to enable UTF8 decoding before importing?

The thing is not UTF vs non-UTF. There are dozens of encodings out there. We have to be able to tackle them all.

damda58:
in the source code of HelloFresh pages we find the charset in the metadata. Is it not possible to recover it here?

```
<html>
<body>
<!--StartFragment-->

<!DOCTYPE html>
--
  | <html lang="fr">
  | <head>
  | <title data-react-helmet="true">Filet mignon de porc avec son jus de cuisson Recette \| HelloFresh</title>
  | <meta data-react-helmet="true" name="viewport" content="width=device-width, initial-scale=1.0"/><meta data-react-helmet="true" http-equiv="Accept-CH" content="DPR,Width,Viewport-Width"/><meta data-react-helmet="true" name="country" content="fr"/>**<meta data-react-helmet="true" name="charset" content="utf-8"/>**

<!--EndFragment-->
</body>
</html>
```

christianlupus:
There is the `meta` tag that contains the information. Then, there is a header (`content-type`) possibly present that might contain the info as well. I do not know which one might have precedence if both are present and do not comply.

christianlupus:
Superseeded bz #1057.

PR #463, opened 2020-12-30
Prevent img loaded event being handled twice

### LINKED ISSUE #462: Error in the JS console in Firefox due to lazy loading of images
**Description**
For some reason, the implementation in the [lazy loading](https://github.com/nextcloud/cookbook/blob/cce3864c79cf21a59a7b02fda34a5739ef2540da/src/components/LazyPicture.vue#L83-L109) seems to cause issues on Firefox.

**Reproduction**
1. Open Firefox
2. Open the web dev tools
3. (Optional) Remove any previous logs
4. Navigate to the start page of cookbook (like `/apps/cookbook`)

One gets messages as below:

grafik

**Expected behavior**
No errors are rosen.

**Browser**
Verified on Firefox 83.0 under Linux and Win10. Chrome seems to be unaffected by this bug.

**Further Information**
It seems that the functions `onThumbnailPreviewLoaded` and `onThumbnailFullyLoaded` are called twice in Firefox. I just added `console.log(*)` entries and found that in Chrome these were called only once (as intended) but twice per image on Firefox.

@seyfeb As this was contributed by you, could you check and verify it is working for you? Any good ideas on how to counteract?

seyfeb:
Closes #462

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,8 @@
   [#447](https://github.com/nextcloud/cookbook/pull/447) @christianlupus
 - Recipe-editing Vue components are not tightly coupled anymore
   [#386](https://github.com/nextcloud/cookbook/pull/386/) @seyfeb
+- Fixed trying to remove already removed img DOM nodes in image lazyloading, fixes #462
+  [#463](https://github.com/nextcloud/cookbook/pull/463/) @seyfeb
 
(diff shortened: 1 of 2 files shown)


christianlupus:
Seems to work well. No more errors in FF or Chrome. Other browsers I have not tested but these can be addressed if the need arises

PR #464, opened 2020-12-30
Warn before leaving recipe editor with changes

### LINKED ISSUE #310: Show warning when navigating away from recipe editing window

seyfeb:
When changes in the recipe form are done by the user, the form is marked dirty. When navigating away from a dirty form, the user is prompted for confirmation.

Closes #310

@sam-19 If you‘d like to share your opinion, you‘re welcome!

### DIFF
--- a/src/components/RecipeEdit.vue
+++ b/src/components/RecipeEdit.vue
@@ -241,6 +241,7 @@ export default {
             })
         },
         save: function() {
+            this.savingRecipe = true
             this.$store.dispatch('setSavingRecipe', { saving: true })
             let $this = this
             if (this.recipe.id) {
@@ -257,6 +258,8 @@ export default {
                 }).fail(function(e) {
                     $this.$store.dispatch('setSavingRecipe', { saving: false })
                     alert(t('cookbook', 'Recipe could not be saved'))
+                }).always(() => {
+                    $this.savingRecipe = false
                 })
             } else {
                 // Create a new recipe
@@ -272,11 +275,12 @@ export default {
                 }).fail(function(e) {
                     $this.$store.dispatch('setSavingRecipe', { saving: false })
                     alert(t('cookbook', 'Recipe could not be saved'))
+                }).always(() => {
+                    $this.savingRecipe = false
                 })
             }
         },
         setup: function() {
-            let $this = this
             this.fetchCategories()
             this.fetchKeywords()
             if (this.$route.params.id) {
@@ -313,7 +317,6 @@ export default {
                     this.allCategories.push(this.recipe['recipeCategory'])
                 }
 
-
                 // Always set the active page last!
                 this.$store.dispatch('setPage', { page: 'edit' })
             } else {
@@ -322,7 +325,7 @@ export default {
             }
             this.recipeInit = JSON.parse(JSON.stringify(this.recipe))
             this.$nextTick(function() {
-                $this.formDirty = false
+                this.formDirty = false
             })
 
         },
@@ -362,7 +365,6 @@ export default {
         this.$root.$off('saveRecipe')
         this.$root.$on('saveRecipe', () => {
             this.save()
-            this.savingRecipe = true
         })
         // Register data load method hook for access from the controls components
         this.$root.$off('reloadRecipeEdit')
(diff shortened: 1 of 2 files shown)


src/components/RecipeEdit.vue
```diff
@@ -129,6 +141,19 @@ export default {
...
+            // Cancel the window unload event
+            e.preventDefault()
+            e.returnValue = ''
+          }   
+        },
+        confirmLeavingPage() {
+          return window.confirm('You have unsaved changes! Do you still want to leave?')
+        },
+        confirmStayInEditedForm() {
+          return !this.savingRecipe && this.formDirty && !this.confirmLeavingPage()
```
christianlupus, line 155: This will continue to new URL when a saving action is just running, right?

src/components/RecipeEdit.vue
```diff
@@ -251,6 +276,7 @@ export default {
             }
         },
         setup: function() {
+            let $this = this
```
christianlupus, line 279: May I ask, mainly out of curiosity, why we need a `$this` variable when we have the `this` one?

src/components/RecipeEdit.vue
```diff
@@ -331,12 +362,17 @@ export default {
         this.$root.$off('saveRecipe')
         this.$root.$on('saveRecipe', () => {
             this.save()
+            this.savingRecipe = true
```
christianlupus, line 365: This will set the saving flag after the saving has been already started/triggered. Maybe we should set it before.

christianlupus (commented):
I am considering if it would be more clear to set the saving flag inside the `save` function. Also, we could reset it on the `done()`/`fail()` functions in order to have a clear idea if an operation is in progress at the moment.

One case I am thinking a bit but that might be off: What happens on a slow network connection when a user clicks save but does not wait for at least the termination of the HTTP request. As the saving flag allows to reload/navigate to another URL, the changes might get lost if the user is too fast. Or am I getting things here wrong?

src/components/RecipeEdit.vue
seyfeb: Good spot :o I introduced it for the $nextTick callback. But actually it’s not necessary since $nextTick is invoked in the `this` context

src/components/RecipeEdit.vue
seyfeb: The idea was: If the user clicked the save button (saving is running), the browser continues to the new URL without checking if the form is dirty (`this.formDirty`) and (if it is dirty) asking for confirmation (`!this.confirmLeavingPage()`)

src/components/RecipeEdit.vue
seyfeb: Yes, I think it’s a good idea to set it in the save function. I’ll update the PR

seyfeb:
> One case I am thinking a bit but that might be off: What happens on a slow network connection when a user clicks save but does not wait for at least the termination of the HTTP request. As the saving flag allows to reload/navigate to another URL, the changes might get lost if the user is too fast. Or am I getting things here wrong?

Actually I’m not sure about this. I tried to provoke this problem but wasn’t successful.

src/components/RecipeEdit.vue
christianlupus: Yes, I see that. This is more related to the overall comment below about the abort of a saving action.

christianlupus:
I tried to provoke this behavior as well but was neither successful. Either one is online and the save is possible or one is offline and one cannot navigate elsewhere. So it should be save to merge this,

Issue #476, opened 2021-01-07
Keywords are not deleted with recipe


seyfeb:
**Description**
When a recipe is deleted the associated keywords should be removed from the database. This doesn’t seem to be the case.

**Reproduction**
1. Create new recipe or edit existing recipe
2. Add a new tag to the recipe (e.g., "testtag")
3. Save recipe
4. Observe: This tag is now available in the multiselect element in the Recipe editor for any recipe and appears in the database (table `oc_cookbook_keywords`).
5. Delete the recipe.
6. Observe: The tag is still available in the multiselect of the recipe editor, even if no recipe with the keyword assigned exists anymore. Also the entry is still present in the database.

**Expected behavior**
Associated keywords should be removed from database when the recipe is deleted

christianlupus:
I will have a look soon.

PR #478, opened 2021-01-07
Remove keywords upon removal of recipe

christianlupus:
Solves #476.


christianlupus:
I yet have to check it's working as expected. Then I can merge.

PR #487, opened 2021-01-09
Create an API endpoint to get the API version

christianlupus:
Especially for 3rd party integrations, it is required to define an API version to clear the possible API endpoints.


christianlupus:
@Teifun2 this should help you get the most recent version of the API

PR #493, opened 2021-01-10
Fixed keywords being returned multiple times for recipes

### LINKED ISSUE #491: Keyword Count does not filter for user in "All Recipes"
**Description**
The node {url}api/recipes returns all keywords without a filter for the current user or a distinct.
With shared recipes this leads to a count of n per recipe instead of 1 for each keyword showing in "All Recipes" 

Unfortunately I have not found the api files to verify in code.

**Reproduction**
In API:
1. Open API Node api/recipes
2. Note the keywords of some recipe
3. Duplicate a keyword entry for that recipe in the database table oc_cookbook_keywords with some fake user_id to simulate a shared recipe
4. Refresh api/recipes and see duplicated keyword

In UI:
1. Go to 'All Recipes'
2. Note the count of some Keyword
3. Duplicate a keyword entry in the database table oc_cookbook_keywords with some fake user_id to simulate a shared recipe
4. Refresh "All Recipes" and see the increased count

**Expected behavior**
Count displays the correct number of recipes / API returns distinct keywords

**Browser**
Firefox

Additional Question: How do those keywords get added for a shared recipe for the other users? For Example UserA adds a Keywod to a recipe, how and when does the keyword get duplicated for UserB? There are recipes in my database where the keywords are not in sync.

seyfeb:
Closes #491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [Unreleased]
 
+### Fixed
+- Fixed keywords of shared recipes counted multiple times, fixes #491
+  [#493](https://github.com/nextcloud/cookbook/pull/493/) @seyfeb
+
 
 ## 0.7.8 - 2021-01-08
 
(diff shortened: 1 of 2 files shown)


christianlupus (approved):
Seems good to me

PR #503, opened 2021-01-12
Allow pasting of normal content

### LINKED ISSUE #475: Single pasted tool/instruction/ingredient is not saved
**Description**
If a single line of tools/instruction/ingredient is pasted (ctrl+v) in the field and the recipe is saved, the value doesn’t get saved

**Reproduction**

1. Copy a string without a newline to the clipboard from somewhere
2. Go to 'Create recipe'
3. Click on '+Add' for either tool/instruction/ingredient
4. Paste the string into the field
5. Save recipe

**Expected behavior**
The pasted string should be saved

**Browser**
Opera, FF

christianlupus:
Closes #475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
   [#501](https://github.com/nextcloud/cookbook/pull/500) @christianlupus
 - Allow to save recipes with custom image URLs
   [#505](https://github.com/nextcloud/cookbook/pull/505) @christianlupus
+- Allow pasting of instructions without newline again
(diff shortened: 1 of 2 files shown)


christianlupus:
@seyfeb would you please share your thoughts on this? I tried to keep things simple.

seyfeb:
I also addressed this while I was editing the file in #497.
My solution was intended to give more flexibility if, in the future, we would like to modify the behavior when pasting. But maybe it’s even better to have a simple solution as you proposed here. 
Anyway, this looks fine and works well for me.

PR #505, opened 2021-01-12
Propagate changes of the image to the recipe

### LINKED ISSUE #492: Pictures are not downloaded
**Description**
When adding a new recipe manually and specifying the URL of the recipe image, the image is not downloaded and the field is not saved.

**Reproduction**
1. Add new recipe manually
2. specify URL for image
3. Save
4. See that the recipe does not have an image
5. Edit the recipe
6. See that the Image field is empty

**Expected behavior**
The image should be downloaded and saved along with the recipe.
Pretty sure this was the behavior of previous versions.

**Screenshots**
n/a

**Browser**
Chrome

christianlupus:
Closes #492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   [#500](https://github.com/nextcloud/cookbook/pull/500) @christianlupus
 - Handle recipes without category well
   [#501](https://github.com/nextcloud/cookbook/pull/500) @christianlupus
+- Allow to save recipes with custom image URLs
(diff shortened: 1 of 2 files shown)


christianlupus:
@seyfeb would you please have a look here as well? Thanks

seyfeb:
looks good from the vue side, but for me it didn’t seem to download and store the image at the url

seyfeb:
Okay, I figured out, that my firewall was blocking the connection. Now it’s downloading the image 👍

PR #511, opened 2021-01-13
Updates color in nutrition info and removes bullet point

### LINKED ISSUE #510: dark Theme Headline nutrition information not really visible
nahrwehrt

seyfeb:
Closes #510

### DIFF
--- a/src/components/RecipeView.vue
+++ b/src/components/RecipeView.vue
@@ -58,7 +58,7 @@
 
                     <section v-if="showNutritions">
                         <h3>{{ t('cookbook', 'Nutrition Information') }}</h3>
-                        <ul>
+                        <ul class="nutrition-items">
                             <recipe-nutrition-info-item v-if="('servingSize' in nutrition) && !isNullOrEmpty(nutrition['servingSize'])" :title="t('cookbook', 'Serving Size')" :data="nutrition['servingSize']" />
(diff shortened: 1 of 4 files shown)


src/components/RecipeView.vue
```diff
@@ -379,6 +379,10 @@ aside {
         margin-top: 2rem;
     }
 
+    .nutrition-items {
+        list-style-type: none !important;
```
christianlupus, line 383: Is this really needed? I am always a bit suspicious when `!important` is in use.

seyfeb: The other ways would be to either 

1. change the definition of `#app-content-wrapper aside ul` in the main `style.css` since this has a higher specificity than using the class
2. use `#app-content-wrapper aside ul.nutrition-items` which might be just as bad as using `!important` :D

For 1. I need to test if there are any unintended consequences when removing the definition. But changing it there might be a sensible way to handle this.

christianlupus: See #515. The global CSS is a bad idea to change the styling of things within vue.

christianlupus: For now, I suggest keeping the `!important` and opening an issue to remind us to come back to it eventually once #515 is finished.

seyfeb: I forgot to push the updated version. We can start deprecating the global style here rather than introducing new code "depending" on the old style file which needs to be changed in the future.

christianlupus:
I am going to merge here. If something broke by this, we might need to tweak it at appropriate locations though.

PR #512, opened 2021-01-14
Update ingredient list visuals

seyfeb:
Depending on the existence of subgroups in the ingredient list (starting with ##) the ingredients are more or less indented. 

The vertical overlap of checkmark and ingredient text, that appeared when there was a line break in the text, is removed.

Also a hanging indent is introduced to make it visually clear when there is a line break within a single ingredient.

Previously:
Screenshot 2021-01-14 at 22 23 38

With subgroup:
Screenshot 2021-01-14 at 22 09 06

Without subgroup:
Screenshot 2021-01-14 at 22 09 45

### DIFF
--- a/src/components/RecipeIngredient.vue
+++ b/src/components/RecipeIngredient.vue
@@ -67,8 +67,8 @@ li {
 
     li:hover > div.checkmark {
         visibility: visible;
-    	opacity: 0.5;
-    	color: var(--color-primary-element);
+        opacity: 0.5;
+        color: var(--color-primary-element);
     }
 
     li > div.ingredient {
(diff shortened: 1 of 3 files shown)


src/components/RecipeIngredient.vue
```diff
@@ -50,19 +50,32 @@ li {
         list-style-type: none;
         font-variant: small-caps;
     }
-    
+
+    li.unindented {
+        position: relative;
+        left: -1.25em;
```
christianlupus, line 56: This can cause trouble. After #515 it might be better to do `margin-left: 0px;` instead.

christianlupus (approved):

PR #513, opened 2021-01-15
Speedup keyword calculation using computed properties

### LINKED ISSUE #489: 0.78 is soo slow
**Description**
My Browser sad, Website is so slow. wait or cancel. Firefox & Chromium. Nextcloud 20.04
0.77 has worked.

christianlupus:
Closes #489

### DIFF
--- a/src/components/AppIndex.vue
+++ b/src/components/AppIndex.vue
@@ -60,12 +60,16 @@ export default {
     },
     data () {
         return {
+            // String-based filters applied to the list
             filters: "",
+            // The known recipes in the cookbook
             recipes: [],
+            // All keyword to filter the recipes for (conjunctively)
             keywordFilter: [],
         }
     },
     computed: {
+        // An array of all keywords in all recipes. These are neither sorted nor unique
         rawKeywords() {
             var keywordArray = this.recipes.map(function(r){
                 if(! 'keywords' in r) {
@@ -79,12 +83,14 @@ export default {
             })
             return [].concat(... keywordArray)
         },
+        // An array of sorted and unique keywords over all the recipes
         uniqKeywords() {
             function uniqFilter(value, index, self) {
                 return self.indexOf(value) === index
             }
             return this.rawKeywords.sort().filter(uniqFilter)
         },
+        // An array of objects that contain the keywords plus a count of recipes asociated with these keywords
         keywordsWithCount() {
             let $this = this
             return this.uniqKeywords.map(function (kw){
@@ -101,12 +107,15 @@ export default {
                 return (k1.name.toLowerCase() > k2.name.toLowerCase()) ? 1 : -1
             })
         },
+        // An array of keyword objects that are currently in use for filtering
         selectedKeywords() {
             return this.keywordsWithCount.filter((kw) => this.keywordFilter.includes(kw.name))
         },
+        // An array of those keyword objects that are currently not in use for filtering
         unselectedKeywords() {
             return this.keywordsWithCount.filter((kw) => ! this.selectedKeywords.includes(kw))
         },
+        // An array of all recipes that are part in all filtered keywords
         recipesFilteredByKeywords() {
             let $this = this
             return this.recipes.filter(function (r) {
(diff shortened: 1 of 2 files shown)


christianlupus:
@seyfeb I have done quite some changes in the file to switch to computed properties. What do you think of this? Do you see any (obvious) bugs?

@poltrian If you like, you can test this version if it solves your issue. For me, it was faster. You can look at the source code artifacts of the checks to get an installable version of the app if you want to give it a try.

src/components/AppIndex.vue
```diff
@@ -39,34 +60,126 @@ export default {
...
+                return (k1.name.toLowerCase() > k2.name.toLowerCase()) ? 1 : -1
+            })
+        },
+        // An array of keyword objects that are currently in use for filtering
+        selectedKeywords() {
+            return this.keywordsWithCount.filter((kw) => this.keywordFilter.includes(kw.name))
+        },
+        // An array of those keyword objects that are currently not in use for filtering
+        unselectedKeywords() {
+            return this.keywordsWithCount.filter((kw) => ! this.selectedKeywords.includes(kw))
```
seyfeb, line 116: Shouldn’t it be faster to do the same (inverse) as for `selectedKeywords`?

```
return this.keywordsWithCount.filter((kw) => ! this.keywordFilter.includes(kw.name))
```

christianlupus: This might be a small bit faster to use the `keywordFilter` instead of `selectedKeywords`. The difference is the `.name` access. Both `keywordFilter` and `selectedKeywords` should have the same number of entries though.

Which one is better readable?

seyfeb: let’s say with your code it’s obvious that no keyword is missing since one partial list is generated by excluding entries which are present in the other partial list

christianlupus: Just tell me what we should do. I'd like to finish this issue soon :smile:.

src/components/AppIndex.vue
```diff
@@ -39,34 +60,126 @@ export default {
...
+        selectableKeywords() {
+            if (this.unselectedKeywords.length === 0) {
+                return []
+            }
+            
+            let $this = this
+            return this.unselectedKeywords.filter(function (kw) {
+                return $this.filteredRecipes.map(function (r) {
+                    return r.keywords && r.keywords.split(',').includes(kw.name)
+                }).reduce((l,r) => l || r, false)
```
seyfeb, line 166: Do you know if this is handled in such a clever way that it already returns if it hits the first `true`?

src/components/AppIndex.vue
```diff
@@ -39,34 +60,126 @@ export default {
...
+            })
+        },
+        // An array of known keywords that are not associated with any visible recipe
+        unavailableKeywords() {
+            return this.unselectedKeywords.filter((kw) => ! this.selectableKeywords.includes(kw))
+        },
+        // An array of recipe objects of all recipes with links to the recipes and a property if the recipe is to be shown 
+        recipeObjects() {
+            let filtered = this.filteredRecipes
```
seyfeb, line 175: if you pass `this` to `map` you don’t need to create a temporary variable, see [the map syntax ref](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

src/components/AppIndex.vue
seyfeb: lol. leave it like this^^

src/components/AppIndex.vue
christianlupus: I do not think so. It first generates the whole bool array and then merges it. The benefit is that the array operation can be parallalized that way (in theory).

seyfeb:
I looked over the code and it looks good. It also worked when I tested it :D

seyfeb (approved):

src/components/AppIndex.vue
christianlupus: Yeah, that makes sense, I was worried that the array would change throughout the invocation. That would render the mapping broken. I will update as you suggested as it seems quite fast in my machine.

seyfeb: Haven’t thought of that.. we should keep that in mind in case somebody experiences some issues

PR #528, opened 2021-01-17
Remove main style file

### LINKED ISSUE #515: Remove pre-vue CSS file
Many styles in the [main css file](https://github.com/nextcloud/cookbook/blob/58a0b9707cf5fd367331c21163dc36826c6d885b/css/style.css#L80-L303) are better off in the corresponding vue files. This way it is quite unclear why what setting was given like it is.

Anything explicitly under `#app`, `#app_content_wrapper` or in general any specific id is most probably related only to a certain view in vue. One (quick) option would maybe be to move these into the corresponding vue file where the id is defined and using a CSS child selector there to reduce the weight and the scope.

seyfeb:
Please test this cautiously since I removed a lot of css stuff that looked unused to me

Closes #515


christianlupus:
@seyfeb I wanted to have a look tomorrow. Today it will be too late. You are ready, I assume?

seyfeb:
Yes and that’s perfectly fine

christianlupus (approved):
I tried to test all cases. I hope i did not forget anything. I found no big issues but a few minor ones. I added a corresponding commit to cover that.

PR #546, opened 2021-01-20
Added Vuex action for updating recipe directory

### LINKED ISSUE #542: false error warning when setting recipes folder
**Description**
When I specify a storage folder for cookbook's data, I am confronted with a **false-positive** error popup

**Reproduction**
1. open cookbook settings
2. choose a folder (shared/not shared, with leading underscore or not, at the root level or deeper)
3. validate
4. popup opens claiming: "Could not set recipe folder to xxx"
5. There is no message from cookbook in the system logs
6. Reload the page, **the folder has actually been selected**, and test-adding recipe does create content in the file folder

**Expected behavior**
since the folder has indeed been selected: show a confirmation popup

**Screenshots**
Screenshot 2021-01-19 111907

**Browser**
Chrome Version 80.0.3987.149 (Official Build) (32-bit) (portableApp)

**Versions**
Nextcloud server version: 20.0.5
Cookbook version: 0.7.10

seyfeb:
Centralized some recipe server call functionality as Vuex actions so they can be easily changed and reused.

- create recipe
- update recipe
- delete recipe
- updating the recipe directory path

Closes #542

### DIFF
--- a/src/components/AppIndex.vue
+++ b/src/components/AppIndex.vue
@@ -18,6 +18,25 @@ export default {
             recipes: []
         }
     },
+    computed: {
+        /**
+         * Is the Cookbook recipe directory currently being changed?
+         */
+        updatingRecipeDirectory () {
+            return this.$store.state.updatingRecipeDirectory
+        }
+    },
+    watch: {
+        /**
+         * If the Cookbook recipe directory currently was changed, reload
+         * the recipes in the index component.
+         */
+        updatingRecipeDirectory(newVal, oldVal) {
+            if (newVal == false && newVal != oldVal) {
+                this.loadAll()
+            }
+        }
+    },
     methods: {
         /**
          * Load all recipes from the database
@@ -39,6 +58,6 @@ export default {
     },
     mounted () {
         this.loadAll()
-    },
+    }
 }
 </script>
--- a/src/components/AppNavi.vue
+++ b/src/components/AppNavi.vue
@@ -354,7 +354,7 @@ export default {
          * Toggle the left navigation pane
          */
         toggleNavigation: function() {
-            this.$store.dispatch('setAppNavigationVisible', { isVisible: !this.$store.state.appNavigationVisible })
+            this.$store.dispatch('setAppNavigationVisible', { isVisible: !this.$store.state.appNavigation.visible })
         },
     },
     mounted () {
--- a/src/components/AppNavi.vue
+++ b/src/components/AppNavi.vue
@@ -294,18 +294,13 @@ export default {
                 t('cookbook', 'Path to your recipe collection'),
                 (path) => {
                     let $this = this
-                    axios({
-                        url: this.$window.baseUrl + '/config',
-                        method: 'POST',
-                        data: { 'folder': path },
-                    })
+                    this.$store.dispatch('updateRecipeDirectory', { dir: path })
                         .then(() => {
-                            $this.loadAll()
-                            .then(() => {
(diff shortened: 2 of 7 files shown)


src/components/AppNavi.vue
```diff
@@ -294,18 +294,13 @@ export default {
...
+                    this.$store.dispatch('updateRecipeDirectory', { dir: path })
                         .then(() => {
-                            $this.loadAll()
-                            .then(() => {
-                                $this.$store.dispatch('setRecipe', { recipe: null })
-                                $this.$window.goTo('/')
-                                $this.recipeFolder = path
-                            })
+                            $this.recipeFolder = path
+                            $this.$root.$emit('refreshNavigation')
```
christianlupus, line 300: Why emitting on the `$root` element?

src/components/AppIndex.vue
```diff
@@ -18,6 +18,25 @@ export default {
...
+        }
+    },
+    watch: {
+        /**
+         * If the Cookbook recipe directory currently was changed, reload
+         * the recipes in the index component.
+         */
+        updatingRecipeDirectory(newVal, oldVal) {
+            if (newVal == false && newVal != oldVal) {
+                this.loadAll()
```
christianlupus, line 36: This will (currently) only load the new recipes. The categories on the left side will not get updated.

christianlupus (changes requested):
When moving back and forth between different folders (without reloading!) I get a message that the categories cannot be fetched. Most probably this is an issue with the synchronization between the global Vuex state and the local list of categories in the AppNavi compnent.

src/components/AppIndex.vue
seyfeb: Yes, this only updates this component (itself). The categories are updated [here](https://github.com/nextcloud/cookbook/blob/b7ac727a100653d4c11ddcca0723a89ab43bb7f0/src/components/AppNavi.vue#L300)

src/components/AppNavi.vue
seyfeb: The `$root` element is used as a global bus here

src/components/AppIndex.vue
christianlupus: I do not know by hard, where the problem is exactly but [here](https://github.com/nextcloud/cookbook/blob/b7ac727a100653d4c11ddcca0723a89ab43bb7f0/src/components/AppNavi.vue#L275) I get the error that `$this.$refs['app-navi-cat-'+i][0]` is undefined when switching back and forth.
Looking at the `$this.$refs` using the JS debugger + Vue dev tools indicated that some of the `app-navi-cat-*` entries are empty. Seems like there is something not synchronized on time.

I have not elaborated further, just saw the error in the JS console and had a quick glance on why this is happening.

seyfeb: I’ll have a look at this. With "switching back and forth" you mean switching between different recipe directories, I guess?

seyfeb: The problem was that the component tried loading recipes for the categories before they were updated in by vue. This was actually already a problem before but couldn’t be notice since updating the recipe folder "failed" anyway ;)
I updated the PR to fix this

src/components/AppNavi.vue
seyfeb: While this is in principal okay, I still changed this a bit

christianlupus (approved):
Now it seems to work and I see no issues ATM.

Issue #537, opened 2021-01-18
Overhaul of the test environment


christianlupus:
The test environment should be updated.

- [ ] Make the `run-local.sh` script command line parameter aware
- [ ] Trigger pulls/builds from docker explicitly
- [ ] Build a docker image to use as a basis for integration tests
- [ ] Better structure the file for better understandability, define useful chunks of functionality
- [ ] Prepare integration testing using docker + phpunit + setup (bash) script
- [ ] [optional] Build a daily/weekly set of prebuilt docker images on the docker hub for faster test setup (only the dependencies)
- [ ] [optional] run the test routines in a docker container in order to have the same commands on GH actions and locally (?)

PR #553, opened 2021-01-23
Extract parsing into its own service class

christianlupus:
This PR should reduce the complexity of the RecipeService and pull out some of the functionality in smaller chunks that might also simplify debugging.

For now, the `recipeParseHtml()` method is only deprecated but should be removed soon.


christianlupus:
@seyfeb I tried here to start the codebase refractory and pull some functionality into its own service. I know this is quite a bigger pack of changes which makes it hard to check line by line. What do you think, how should we proceed here? First writing tests for all changes involved here or building the tests later? I tried to make it possible to test the code but did not write the tests yet.

If we want to do this with everything tested, I would have to solve #537 first, I assume.

Any further remarks?

seyfeb:
I’m not sure either. We might get some regressions when doing this without tests, so this might be a good moment to start writing tests. However, I’m not sure if, with our limited time to work on this, it might just be okay to deal with them later and first get the better structure in the code and hope for the best (keeping in mind to add tests once we have #537).

christianlupus:
OK, then we wait for #537, rebase to have these changes, and try to implement some tests afterward. I converted to draft to make this obvious.

christianlupus:
Closes #740

christianlupus:
OK, I added tests for all new classes except for the `RecipeService`. This special class is not yet refactored and is not in a testable shape.

Also, no integration tests were written so far. This is related to the fact that no clear idea of what sites should be taken as examples for these tests.

lib/Helper/HTMLFilter/AbstractHtmlFilter.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace OCA\Cookbook\Helper\HTMLFilter;
+
+abstract class AbstractHtmlFilter {
+	
+	/**
+	 * Filter the HTML according to the rules of this class
+	 *
+	 * This class operates on the original HTML code as passed by reference.
```
seyfeb, line 10:
```suggestion
	 * This class operates on the original HTML code as passed by reference and may therefore modify the HTML string.
```

lib/Helper/HTMLParser/HttpJsonLdParser.php
```diff
@@ -0,0 +1,189 @@
+<?php
+
+namespace OCA\Cookbook\Helper\HTMLParser;
+
+use OCA\Cookbook\Exception\HtmlParsingException;
+use OCP\IL10N;
+use OCA\Cookbook\Service\JsonService;
+
+/**
+ * This class is an AbsractHtmlParser that tries to extract a JSON+LD script from the HTML page.
```
seyfeb, line 10:
```suggestion
 * This class is an AbstractHtmlParser which tries to extract a JSON+LD script from the HTML page.
```

lib/Helper/HTMLParser/HttpJsonLdParser.php
```diff
@@ -0,0 +1,189 @@
...
+		
+		$json_ld_elements = $xpath->query("//*[@type='application/ld+json']");
+		
+		foreach ($json_ld_elements as $json_ld_element) {
+			if (!$json_ld_element || !$json_ld_element->nodeValue) {
+				continue;
+			}
+			
+			try {
+				return $this->parseJsonLdElement($json_ld_element);
```
seyfeb, line 37: This takes the first element it finds and returns it. What if there are multiple valid ones?

lib/Helper/HTMLParser/HttpJsonLdParser.php
```diff
@@ -0,0 +1,189 @@
...
+	 *
+	 * @param array $json The JSON object to check
+	 */
+	private function mapGraphField(array &$json) {
+		if (isset($json['@graph']) && is_array($json['@graph'])) {
+			$tmp = $this->searchForRecipeInArray($json['@graph']);
+			
+			if ($tmp !== null) {
+				$json = $tmp;
+				return;
```
seyfeb, line 127:
```suggestion
```

lib/Helper/HTMLParser/HttpJsonLdParser.php
```diff
@@ -0,0 +1,189 @@
...
+	 * input parameter.
+	 * @param array $json The JSON object to insprect
+	 */
+	private function mapArray(array &$json) {
+		if (isset($json[0])) {
+			$tmp = $this->searchForRecipeInArray($json);
+			
+			if ($tmp !== null) {
+				$json = $tmp;
+				return;
```
seyfeb, line 148:
```suggestion
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
+<?php
+
+namespace OCA\Cookbook\Helper\HTMLParser;
+
+use OCA\Cookbook\Exception\HtmlParsingException;
+use OCP\IL10N;
+
+/**
+ * This class is an AbsractHtmlParser that tries to extract microdata from the HTML page.
```
seyfeb, line 9:
```suggestion
 * This class is an AbstractHtmlParser which tries to extract microdata from the HTML page.
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+	/**
+	 * Parse a DOM node that represents a recipe
+	 *
+	 * @param \DOMNode $recipeNode The DOM node to parse
+	 */
+	private function parseRecipe(\DOMNode $recipeNode) {
+		$this->seachSimpleProperties($recipeNode, 'name');
+		$this->seachSimpleProperties($recipeNode, 'keywords');
+		$this->seachSimpleProperties($recipeNode, 'category');
+		$this->seachSimpleProperties($recipeNode, 'recipeYield');
```
seyfeb, line 62:
```suggestion
		$this->searchSimpleProperties($recipeNode, 'name');
		$this->searchSimpleProperties($recipeNode, 'keywords');
		$this->searchSimpleProperties($recipeNode, 'category');
		$this->searchSimpleProperties($recipeNode, 'recipeYield');
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+		
+		$this->parseImage($recipeNode);
+		$this->parseIngredients($recipeNode);
+		$this->parseInstructions($recipeNode);
+		
+		$this->fixupInstructions($recipeNode);
+	}
+	
+	/**
+	 * Make one final desparate attempt at getting the instructions
```
seyfeb, line 72:
```suggestion
	 * Make one final desperate attempt at getting the instructions
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+		return $this->searchMultipleProperties($recipeNode,
+			['recipeInstructions', 'instructions', 'steps', 'guide'],
+			['content'],
+			'recipeInstructions');
+	}
+	
+	/**
+	 * Search for microdata properties under various names in the recipe and save as a list
+	 *
+	 * If a property can be named differntly and found under differnt property names within
```
seyfeb, line 131:
```suggestion
	 * If a property can be named differently and found under different property names within
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+	 * should still be backward compatible.
+	 *
+	 * @param \DOMNode $recipeNode The node of the recipe to look for properties in
+	 * @param array $properties The properties to look for one-by-one
+	 * @param array $attributes The attributes that will contain the data
+	 * @param string $dst The name of the property list in the internal structure
+	 * @return boolean true, if the property was found
+	 */
+	private function searchMultipleProperties(\DOMNode $recipeNode, array $properties,
+		array $attributes, string $dst) {
```
seyfeb, line 144:
```suggestion
		array $attributes, string $dst): bool {
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+	 * Search for microdata properties of the recipe.
+	 *
+	 * Within the recipe node search for DOM entries that represent a property and assign the
+	 * internal structure to their value.
+	 *
+	 * @param \DOMNode $recipeNode The node of the recipe under which to look for properties
+	 * @param string $property The property name to look for
+	 * @return bool true, if the property was found
+	 */
+	private function seachSimpleProperties(\DOMNode $recipeNode, string $property): bool {
```
seyfeb, line 174:
```suggestion
	private function searchSimpleProperties(\DOMNode $recipeNode, string $property): bool {
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+			
+			if (count($values) === 1) {
+				$values = $values[0];
+			}
+		} catch (AttributeNotFoundException $ex) {
+			return false;
+		}
+		
+		$this->recipe[$property] = $values;
+		return true;
```
seyfeb, line 188: Can this be replaced by a call to `searchMultipleProperties`?

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+	 * @param \DOMNode $recipeNode The root node to search in
+	 * @param string $prop The name of the property to look for
+	 * @return \DOMNodeList A list of all found child nodes with the given property
+	 */
+	private function searchChildEntries(\DOMNode $recipeNode, string $prop): \DOMNodeList {
+		return $this->xpath->query("//*[@itemprop='$prop']", $recipeNode);
+	}
+	
+	/**
+	 * Extract the value from a HTML attribute
```
seyfeb, line 203:
```suggestion
	 * Extract the value from an HTML attribute
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+	 *
+	 * This method checks a set of notes if any of these nodes contain an attribute that can be used
+	 * to extract some microdata. The nodes are iterated one-by-one. As soon as a match is found,
+	 * the method retunes the corresponding value and terminates.
```
seyfeb, line 207:
```suggestion
	 * the method returns the corresponding value and terminates.
```

lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -0,0 +1,265 @@
...
+	 *
+	 * In each node each attribute is checked (in order of occurence) if the node has such an attribute.
```
seyfeb, line 209:
```suggestion
	 * In each node each attribute is checked (in order of occurrence) if the node has such an attribute.
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+namespace OCA\Cookbook\Helper;
+
+use DOMDocument;
+use OCP\IL10N;
+use LibXMLError;
+use OCP\ILogger;
+use OCA\Cookbook\Exception\ImportException;
+
+/**
+ * This class allows to parse a HTML document into a DOMDocuemnt.
```
seyfeb, line 12:
```suggestion
 * This class allows to parse an HTML document into a DOMDocument.
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+	 */
+	private $l;
+
+	public function __construct(ILogger $logger, IL10N $il10n) {
+		$this->logger = $logger;
+		$this->l = $il10n;
+	}
+
+	/**
+	 * Parse a HTML file logging any error in the syntax
```
seyfeb, line 59:
```suggestion
	 * Parse an HTML file logging any error in the syntax
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+	 *
+	 * @param DOMDocument $dom The dom docuemnt to use. Just give a new document in doubt.
```
seyfeb, line 61:
```suggestion
	 * @param DOMDocument $dom The dom document to use. Just give a new document in doubt.
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	/**
+	 * Get the error state of the last parsing routine.
+	 *
+	 * See the constants in the class for possible values.
+	 *
+	 * @return int
+	 */
+	public function getState() {
```
seyfeb, line 95:
```suggestion
	public function getState(): int {
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+
+	/**
+	 * Compress the xml errors to small mount of log messages
+	 *
+	 * The return value indicates what the most critical value was.
+	 * The value can be PARSING_SUCCESS, PARSING_WARNING, PARSING_ERROR, or PARSING_FATAL_ERROR.
+	 *
+	 * @param array $errors The array of all parsed errors
+	 * @param string $url The parsed URL
+	 * @return int Indicator what the most severe issue was
```
seyfeb, line 107:
```suggestion
	 * @return int Indicator what the most severe issue was
	 * @throws \Exception
```

Documented return type does not match void return type

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+	 *
+	 * This will iterate over all errors and count for each error code how often the code is present in the errors.
+	 * Also the first error of each code will be stored.
+	 *
+	 * The returned array will contain an entry for each found error code.
+	 * The error code is the corresponding key in the array.
+	 *
+	 * Each entry in the array is itself an array containing the following entries:
+	 * First, there is the key `count` that contains the number of occurences of the corresponding error code.
+	 * Second, there is the key `first` thst contains the first error that was found with this error code.
```
seyfeb, line 128:
```suggestion
	 * First, there is the key `count` that contains the number of occurrences of the corresponding error code.
	 * Second, there is the key `first` that contains the first error that was found with this error code.
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+	}
+	
+	/**
+	 * Log the found error groups to the NC core error logger.
+	 *
+	 * As a side effect, the property $state will be updated to the most severe error found.
+	 *
+	 * @param array $groupedErrors The grouped errors as defined in groupErrors
+	 * @param string $url The URL to import
+	 * @return void
```
seyfeb, line 162:
```suggestion
	 * @return void
	 * @throws \Exception
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+	 */
+	private function logAllErrors(array $groupedErrors, string $url) {
```
seyfeb, line 164:
```suggestion
	private function logAllErrors(array $groupedErrors, string $url): void {
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+				case LIBXML_ERR_FATAL:
+					$this->logFatalError($code, $group, $url);
+					break;
+				default:
+				throw new \Exception($this->l->t('Unsupported error level during parsing of XML output.'));
+			}
+		}
+	}
+
+	private function logWarning(int $code, array $group, string $url) {
```
seyfeb, line 182:
```suggestion
	private function logWarning(int $code, array $group, string $url): void {
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+		$msg = $this->l->t('Warning %u occurred %u times while parsing %s.', [$code, $group['count'], $url]);
+		$this->logger->notice($this->formatError($msg, $group['first']));
+		$this->state = max($this->state, self::PARSING_WARNING);
+	}
+
+	
+	private function logError(int $code, array $group, string $url) {
```
seyfeb, line 189:
```suggestion
	private function logError(int $code, array $group, string $url): void {
```

lib/Helper/HtmlToDomParser.php
```diff
@@ -0,0 +1,206 @@
...
+		$msg = $this->l->t('Error %u occurred %u times while parsing %s.', [$code, $group['count'], $url]);
+		$this->logger->warning($this->formatError($msg, $group['first']));
+		$this->state = max($this->state, self::PARSING_ERROR);
+	}
+
+	private function logFatalError(int $code, array $group, string $url) {
```
seyfeb, line 195:
```suggestion
	private function logFatalError(int $code, array $group, string $url): void {
```

lib/Service/HtmlDownloadService.php
```diff
@@ -0,0 +1,110 @@
...
+	}
+	
+	/**
+	 * Download a recipe URL and extract the JSON from it
+	 *
+	 * The return value is one of the constants of the class PARSE_SUCCESS, PARSE_WARNING,
+	 * PARSE_ERROR, or PARSE_FATAL_ERROR.
+	 *
+	 * @param int The indicator if the HTML page was correctly parsed
+	 * @return array The included JSON data array, unfiltered
```
seyfeb, line 55: Wrong param and return type

lib/Service/HtmlDownloadService.php
```diff
@@ -0,0 +1,110 @@
...
+		$dom = new DOMDocument();
+		$this->dom = $this->htmlParser->loadHtmlString($dom, $url, $html);
+		return $this->htmlParser->getState();
+	}
+	
+	/**
+	 * Get the HTML docuemnt after it has been downloaded and parsed with downloadRecipe()
+	 * @return \DOMDocument The loaded HTML document or null if document could not be loaded successfully
+	 */
+	public function getDom(): ?\DOMDocument {
```
seyfeb, line 76:
```suggestion
	 * @return ?DOMDocument The loaded HTML document or null if document could not be loaded successfully
	 */
	public function getDom(): ?DOMDocument {
```

lib/Service/HtmlDownloadService.php
```diff
@@ -0,0 +1,110 @@
...
+		return $this->dom;
+	}
+
+	/**
+	 * Fetch a HTML page from the internet
```
seyfeb, line 81:
```suggestion
	 * Fetch an HTML page from the internet
```

lib/Service/RecipeService.php
```diff
@@ -796,36 +813,27 @@ public function addRecipe($json) {
...
 		}
-
-		$json = $this->parseRecipeHtml($url, $html);
-
+		
+		$json = $this->checkRecipe($json);
+		
 		if (!$json) {
-			throw new Exception('No recipe data found');
+			$this->logger->error('Importing parsers resulted in null recipe. ' .
```
seyfeb, line 834:
```suggestion
			$this->logger->error('Importing parsers resulted in null recipe.' .
```

seyfeb (commented):
I tested this with a recipe and it worked as expected. I did not go through all the test cases though..

seyfeb:
> Also, no integration tests were written so far. This is related to the fact that no clear idea of what sites should be taken as examples for these tests.

This is especially difficult since third-party pages may change in the future and would therefore not be suited for tests...

lib/Helper/HTMLParser/HttpJsonLdParser.php
christianlupus: Are you talking of different Recipe objects within a single page or different schema.org objects?

The `parseJsonLdElement` will throw an Exception if no recipe is found. So, the case of other schema objects can be neglected as far as I see.

Regarding the multi-recipe pages: I have no clue, what we could do to handle this case better. In fact, it is the user's decision what recipe he wants to import. So, he is responsible for checking if the correct recipe was imported.

seyfeb: I meant multiple json ld elements. Best case (most effort) scenario would probably be to parse on client device and prompt the user with a selection to choose from. However, since this is really an edge case I vote for ignoring it until there actually is someone with that use case.

lib/Helper/HTMLParser/HttpMicrodataParser.php
christianlupus: Should be done.

christianlupus:
I am going to merge this now. This will allow breaking the work into smaller chunks that can be handled more easily. Especially adding more test and integration tests can be done separately. See also #1017.

PR #555, opened 2021-01-23
Renaming category

### LINKED ISSUE #322: Batch category renaming
## Current state

Currently recipes are put in a category by typing in the category name. The categories are then created by clustering recipes with the same category string. Renaming a category, e.g., from `Starters` to `Starters & Antipasti` would require manually opening each recipe of the category and updating the category field.

## Proposed state

Add some editing UI (e.g., a pencil next to the category name that opens a popup for renaming) and iterate over all recipes matching the old category string and update to the new string.

Slightly related to #313.

seyfeb:
- added API endpoint (`'url' => '/api/category/{category}', 'verb' => 'PUT'`)
- made categories in navigation pane editable

Closes #322

### DIFF
--- a/lib/Controller/MainController.php
+++ b/lib/Controller/MainController.php
@@ -238,6 +238,8 @@ class MainController extends Controller {
 				$r['recipeCategory'] = $json['name'];
 				$this->service->addRecipe($r);
 			}
+			// Update cache
+			$this->dbCacheService->updateCache();
 
 			return new DataResponse($json['name'], Http::STATUS_OK, ['Content-Type' => 'application/json']);
 		} catch (\Exception $e) {
--- a/lib/Controller/MainController.php
+++ b/lib/Controller/MainController.php
@@ -227,13 +227,13 @@ class MainController extends Controller {
 
 		$json = $this->restParser->getParameters();
 		if (!$json || !isset($json['name']) || !$json['name']) {
-			throw new Exception('New category name not found');
+			return new DataResponse('New category name not found in data', 400);
 		}
 
 		$category = urldecode($category);
 		try {
 			$recipes = $this->service->getRecipesByCategory($category);
-			foreach ($recipes as $i => $recipe) {
+			foreach ($recipes as $recipe) {
 				$r = $this->service->getRecipeById($recipe['recipe_id']);
 				$r['recipeCategory'] = $json['name'];
 				$this->service->addRecipe($r);
@@ -241,7 +241,6 @@ class MainController extends Controller {
 
 			return new DataResponse($json['name'], Http::STATUS_OK, ['Content-Type' => 'application/json']);
 		} catch (\Exception $e) {
-			error_log($e);
 			return new DataResponse($e->getMessage(), 500);
 		}
 	}
--- a/lib/Controller/MainController.php
+++ b/lib/Controller/MainController.php
@@ -222,6 +222,34 @@ class MainController extends Controller {
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 */
+	public function categoryUpdate($category) {
+		$this->dbCacheService->triggerCheck();
+
+		$json = $this->restParser->getParameters();
+		if (!$json || !isset($json['name']) || !$json['name']) {
+			throw new Exception('New category name not found');
+		}
+
+		$category = urldecode($category);
+		try {
+			$recipes = $this->service->getRecipesByCategory($category);
+			foreach ($recipes as $i => $recipe) {
+				$r = $this->service->getRecipeById($recipe['recipe_id']);
+				$r['recipeCategory'] = $json['name'];
+				$this->service->addRecipe($r);
+			}
+
+			return new DataResponse($json['name'], Http::STATUS_OK, ['Content-Type' => 'application/json']);
+		} catch (\Exception $e) {
+			error_log($e);
+			return new DataResponse($e->getMessage(), 500);
+		}
+	}
+
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 */
 	public function tags($keywords) {
 		$this->dbCacheService->triggerCheck();
 		$keywords = urldecode($keywords);
--- a/lib/Controller/MainController.php
+++ b/lib/Controller/MainController.php
@@ -73,7 +73,7 @@ class MainController extends Controller {
 			'cookbook_version' => [0, 7, 10], /* VERSION_TAG do not change this line manually */
 			'api_version' => [
 				'major' => 0,
-				'minor' => 1
+				'minor' => 2
 			]
 		];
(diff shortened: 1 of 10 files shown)


lib/Controller/MainController.php
```diff
@@ -218,6 +218,34 @@ public function category($category) {
...
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 */
+	public function categoryUpdate($category) {
+		$this->dbCacheService->triggerCheck();
+
+		$json = $this->restParser->getParameters();
+		if (!$json || !isset($json['name']) || !$json['name']) {
+			throw new Exception('New category name not found');
```
christianlupus, line 230: Throwing a generic Exception is hard to catch later on. Also this will trigger a default request answer by NC that contains a stack trace and a human-readable HTML page to debug. Not ideal for AJAX requests.

As we are in the Controller here, I suggest creating a 4xx DataResponse here instead.

If you want to stick with the generic exception here, you should make it `\Exception` as we are in a namespace.

lib/Controller/MainController.php
```diff
@@ -218,6 +218,34 @@ public function category($category) {
...
+		}
+
+		$category = urldecode($category);
+		try {
+			$recipes = $this->service->getRecipesByCategory($category);
+			foreach ($recipes as $i => $recipe) {
```
christianlupus, line 236: The `$i` is not needed here.

lib/Controller/MainController.php
```diff
@@ -218,6 +218,34 @@ public function category($category) {
...
+				$r = $this->service->getRecipeById($recipe['recipe_id']);
+				$r['recipeCategory'] = $json['name'];
+				$this->service->addRecipe($r);
+			}
+
+			return new DataResponse($json['name'], Http::STATUS_OK, ['Content-Type' => 'application/json']);
+		} catch (\Exception $e) {
+			error_log($e);
```
christianlupus, line 244: Logging to the system log might be better off with the NC logfile, don't you think?

src/components/AppNavi.vue
```diff
@@ -382,9 +411,10 @@ export default {
     background-repeat: no-repeat;
 }
 
->>> .app-navigation-entry *:not(.app-navigation-entry-icon) {
+/* >>> .app-navigation-entry *:not(.app-navigation-entry-icon) {
     background: initial !important;
-}
+} */
```
christianlupus, line 416: Remove it instead of commenting

src/components/RecipeEdit.vue
```diff
@@ -461,6 +471,19 @@ beforeRouteLeave (to, from, next) {
     padding: 1rem;
 }
 
+.overlay {
+    display: block;
+    position: absolute;
+    width: 100%;
+    height: 100%;
```
christianlupus, line 478:
```suggestion
    height: 100%;
    left: 0px;
    top: 0px;
```

This should avoid intermediate showing of overflow bars of the browser.

christianlupus (commented):
I found an issue with the code changes when you change the name multiple times. Do you want to fix this before merging or want me to open a new issue afterward to keep things separated?

christianlupus:
One more thing: ATM I have to reject the PR. We need to update the [API minor version](https://github.com/nextcloud/cookbook/blob/76bd3116e9266dd523f4dfed44935e6b7034e9ac/lib/Controller/MainController.php#L74-L77) as we have some changes of the API endpoints.

lib/Controller/MainController.php
seyfeb: I deleted it

seyfeb: removed

lib/Controller/MainController.php
seyfeb: updated to returning 400

lib/Controller/MainController.php
seyfeb: removed

src/components/AppNavi.vue
seyfeb: removed

src/components/RecipeEdit.vue
seyfeb: added

seyfeb:
> I found an issue with the code changes when you change the name multiple times. Do you want to fix this before merging or want me to open a new issue afterward to keep things separated?

I did quite some changes since your comment. @christianlupus Does this still exist?

> One more thing: ATM I have to reject the PR. We need to update the API minor version as we have some changes of the API endpoints.

Okay, we could also create a separate controller for categories.

christianlupus:
> > I found an issue with the code changes when you change the name multiple times. Do you want to fix this before merging or want me to open a new issue afterward to keep things separated?
> 
> I did quite some changes since your comment. @christianlupus Does this still exist?

Not exactly the same but similar. As an image says more than 1000 words and a video has 25 images per sec, this is worth a whole book :smile:: 

https://user-images.githubusercontent.com/8202922/105676772-456c8f00-5eeb-11eb-91c6-d76a49491266.mp4

> > One more thing: ATM I have to reject the PR. We need to update the API minor version as we have some changes of the API endpoints.
> 
> Okay, we could also create a separate controller for categories.

No, keeping that in the main controller is OK so far (no talked from a codebase refactoring perspective).
What I mean is the fact that 3rd part apps like the one from @Teifun2 use our API endpoints as REST service. To allow them to have a clear understanding, which cookbook app version is running, I introduced #487.
The cookbook version is incremented automatically by the bumping script but the API version must be handled manually. I just pushed a commit to update to API version 0.2.

seyfeb:
Alright, this was really helpful. After updating some more stuff, I think this problem should be fixed and I was able to do some other improvements as well.

GutHib:
Seems like you guys can read my thoughts. Nice addition - thank you.

christianlupus:
I just had a look and it seems to work "mostly". I have one issue but it seems this is an upstream bug of the nextcloud-vue components: Open the edit field and try to move the cursor by clicking with the mouse. In my instance, this triggered the `:to` link to be navigated. If @seyfeb has no idea why this is happening, I think we can merge this PR. I will just rebase to get rid of the conflicts.

seyfeb:
> I just had a look and it seems to work "mostly". I have one issue but it seems this is an upstream bug of the nextcloud-vue components: Open the edit field and try to move the cursor by clicking with the mouse. In my instance, this triggered the `:to` link to be navigated. If @seyfeb has no idea why this is happening, I think we can merge this PR. I will just rebase to get rid of the conflicts.

We might want to consider updating the `@nextcloud/vue` dependency and see if this issue is gone. However this is quite a version jump so we need to carefully check the breaking changes of the versions in between.

christianlupus:
I am going to merge this now and open an issue that might get closed once the problem is identified and solved.

PR #562, opened 2021-01-28
Reference recipes

### LINKED ISSUE #209: Link to other recipes with hashes
Allow linking of a recipe as a dependency for another recipe.
eg. There are several basic dough which are used for different pastries

- [ ] May use `#recipes/2443839` as shortcode for a cooking step in edit mode.
  - [ ] If a # is inserted in edit mode show a list of recipe for inserting.
  - [ ] Show last used recipe for inserting first.
- [ ] Show the recipe steps inside and add ingredients to the list in the main recipe.

seyfeb:
Added functionality to reference other recipes by id in description, tools, ingredients, and instructions

- In recipe editor, when entering '#' a popup with a multiselect is shown that allows selecting a reference.
- the popup is shown in fullscreen for small window widths
- When selecting a recipe a reference is entered: `#r/123`
- in recipe viewer this is converted to `#123` linking to the recipe page

Closes #209

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@
   [#381](https://github.com/nextcloud/cookbook/pull/381) @thembeat
 - Changing category name for all recipes in a category
   [#555](https://github.com/nextcloud/cookbook/pull/555/) @seyfeb
-  
+- Functionality to reference other recipes by id in description, tools, ingredients, and instructions
+  [#562](https://github.com/nextcloud/cookbook/pull/562/) @seyfeb
+
 ### Changed
 - Using computed property in recipe view
   [#522](https://github.com/nextcloud/cookbook/pull/522/) @seyfeb
--- a/src/components/RecipeEdit.vue
+++ b/src/components/RecipeEdit.vue
@@ -2,7 +2,7 @@
     <div class="wrapper">
         <div class="overlay" :class="{hidden: !overlayVisible}" />
         <EditInputField :fieldType="'text'" :fieldLabel="t('cookbook', 'Name')" v-model="recipe['name']" />
-        <EditInputField :fieldType="'markdown'" :fieldLabel="t('cookbook', 'Description')" v-model="recipe['description']" />
+        <EditInputField :fieldType="'markdown'" :fieldLabel="t('cookbook', 'Description')" v-model="recipe['description']" :referencePopupEnabled="true" />
         <EditInputField :fieldType="'url'" :fieldLabel="t('cookbook', 'URL')" v-model="recipe['url']" />
         <EditImageField :fieldLabel="t('cookbook', 'Image')" v-model="recipe['image']" />
         <EditTimeField :fieldLabel="t('cookbook', 'Preparation time')" v-model="prepTime" />
@@ -15,6 +15,7 @@
         <EditInputGroup :fieldName="'tool'" :fieldType="'text'" :fieldLabel="t('cookbook', 'Tools')"  v-model="recipe['tool']" v-bind:createFieldsOnNewlines="true" />
         <EditInputGroup :fieldName="'recipeIngredient'" :fieldType="'text'" :fieldLabel="t('cookbook', 'Ingredients')" v-model="recipe['recipeIngredient']" v-bind:createFieldsOnNewlines="true" />
         <EditInputGroup :fieldName="'recipeInstructions'" :fieldType="'textarea'" :fieldLabel="t('cookbook', 'Instructions')"  v-model="recipe['recipeInstructions']" v-bind:createFieldsOnNewlines="true" v-bind:showStepNumber="true" />
+        <edit-multiselect-popup ref="referencesPopup" class="referencesPopup" :class="{visible: referencesPopupFocused}" :options="allrecipeOptions" track-by="recipe_id" label="name" :loading="loadingRecipeReferences" :focused="referencesPopupFocused" />
     </div>
 </template>
 
@@ -26,6 +27,7 @@ import EditInputField from './EditInputField'
 import EditInputGroup from './EditInputGroup'
 import EditMultiselect from './EditMultiselect'
 import EditMultiselectInputGroup from './EditMultiselectInputGroup'
+import EditMultiselectPopup from './EditMultiselectPopup'
 import EditTimeField from './EditTimeField'
 
 export default {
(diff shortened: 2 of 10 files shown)


christianlupus:
I will have to play with the PR a bit. There are quite some changes so I cannot oversee statically fi this will work.

I have two points seen during a quick look at the code:
1. Usage if `v-html` must be used carefully. This allows CSRF attacks if corresponding HTML code is embedded in the corresponding texts. This must be carefully checked and eventually escaped.
2. This will conflict heavily with #381.

seyfeb:
> I will have to play with the PR a bit. There are quite some changes so I cannot oversee statically fi this will work.

Definitely. The main change (in lines of code) is certainly the new component `EditMultiselectPopup.vue`. In the input fields, registering the `#` press and pasting the reference is done. In the `RecipeView` component the references are converted to links.

> 1. Usage if `v-html` must be used carefully. This allows CSRF attacks if corresponding HTML code is embedded in the corresponding texts. This must be carefully checked and eventually escaped.

That is a good point. As Vue does this automatically in most cases I sometimes forget about it. I add some escaping mechanism before inserting it there.

> 2. This will conflict heavily with #381.

It conflicts, but I hope™ that it’s easily fixable.

seyfeb:
I merged the Markdown support and rebased / extended this PR on this basis. It should work for both markdown and non-markdown input fields now!

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 ## [Unreleased]
 
 ### Added
+<<<<<<< HEAD
```
christianlupus, line 4:
```suggestion
```

src/components/RecipeEdit.vue
```diff
@@ -89,7 +93,18 @@ export default {
...
+                { key: 'unsaturatedFatContent', label: t('cookbook', 'Unsaturated-fat content'), placeholder: t('cookbook', 'E.g.: 2 g (amount & unit)') }],
+            referencesPopupFocused: false,
+            popupContext: undefined,
+            loadingRecipeReferences: true
+        }
+    },
+    computed: {
+        allrecipeOptions() {
+            return this.allRecipes.map(r => {
+                    return {recipe_id: r.recipe_id, name: r.recipe_id + ": " + r.name}
```
christianlupus, line 105: Maybe renaming to `title` instead of `name` might be favorable to have an open to extend later.

src/components/RecipeEdit.vue
```diff
@@ -373,7 +390,39 @@ export default {
...
+            .then(function (response) {
+                $this.allRecipes = response.data
+            })
+            .catch(function (e) {
+                console.log(e)
+            })
+            .then(() => {
+                // finally
+                $this.loadingRecipeReferences = false
+            })
```
christianlupus, line 425: This is something we have in the global Vuex state, don't we?

src/components/RecipeView.vue
```diff
@@ -223,6 +239,20 @@ export default {
...
+            return unsafeString
+                .replace(/&/g, "&amp;")
+                .replace(/\~/g, "&#732;")
+                .replace(/</g, "&lt;")
+                .replace(/>/g, "&gt;")
+                .replace(/"/g, "&quot;")
+                .replace(/'/g, "&#039;");
+        },
+        convertRecipeReferences: function(text) {
+            let re = /(^|\s)#r\/(\d+)(\s|$)/g
```
christianlupus, line 252: This will break when two links are just adjacent (aka `#r/123 #r/124`). Here a [look-ahead](https://www.stefanjudis.com/today-i-learned/the-complicated-syntax-of-lookaheads-in-javascript-regular-expressions/) might be needed.

christianlupus (changes requested):
There is one issue with the Changelog that must be tackled and a few minor things that can be taken care of but are mainly border cases.

Apart from these it seems to work and can be merged once these are ironed out.

src/components/RecipeEdit.vue
seyfeb: sounds sensible

src/components/RecipeEdit.vue
seyfeb: We don’t, but I when implementing this I was wondering if we should. Then we wouldn’t have to fetch the data every time we open the editor. On the other hand, if we fetch it every time, we can be sure that the data is up to data (e.g. if someone else added a new recipe).

src/components/RecipeView.vue
seyfeb: I’m really struggling with this 😓 

This won’t work
`(?<=^|\s)(^|\s)#r\/(\d+)(?=$|\s)(\s|$)`

This detects the occurrences but doesn’t capture the first space (if present) and breaks newlines
`(?<=^|\s)#r\/(\d+)(?=$|\s)(\s|$)`

any ideas @christianlupus?

seyfeb: Actually, the second one works. I forgot that Markdown needs an empty line before it renders a line break 😄

src/components/RecipeEdit.vue
christianlupus: Ahh, we have only the current recipe there. My bad, sorry. I would say we keep it like this. Doing more caching might be later an option if performance gets troublesome.

src/components/RecipeView.vue
```diff
@@ -223,6 +239,20 @@ export default {
...
+            return unsafeString
+                .replace(/&/g, "&amp;")
+                .replace(/\~/g, "&#732;")
+                .replace(/</g, "&lt;")
+                .replace(/>/g, "&gt;")
+                .replace(/"/g, "&quot;")
+                .replace(/'/g, "&#039;");
+        },
+        convertRecipeReferences: function(text) {
+            let re = /(?<=^|\s)#r\/(\d+)(?=$|\s)(\s|$)/g
```
christianlupus, line 252:
```suggestion
            let re = /(?<=^|\s|[,._+&?!-])#r\/(\d+)(?=$|\s|[.,_+&?!-])/g
```

This sounds still strange to me (a regular match after a lookahead does not make much sense to me). I find https://regex101.com very useful for testing. I added some punctuation before and after to allow matching with them adjacent as well.

christianlupus (approved):
You might want to have a look at the minor change I suggested but apart from that I think, this can be merged.

src/components/RecipeView.vue
seyfeb: okay, your probably right. that’s probably just duplicating things. however one needs to edit the replacement statement as well, since no second captured group exists anymore.

PR #580, opened 2021-02-13
Deprecate some routes by comments

christianlupus:
We have still some old API endpoints that should be cleaned up. This PR should clean up no longer used routes and structure the code better.


christianlupus:
@Teifun2 please look if you are using one of the `Unknown usage` endpoints. These might still work currently but if we are deciding to deprecate them (I assume so), this might cause issues for you. There are alternatives under the `/api/recipes` tree.

PR #605, opened 2021-02-22
adjust to new dbal

icewind1991:
adjust database code to new dbal shipped with nc21.

Fixes

```
Exception: Undefined class constant 'STRING' at lib/private/AppFramework/Http/Dispatcher.php line 159
```

error when trying to use with nc21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
   [#620](https://github.com/nextcloud/cookbook/pull/620) @christianlupus
 - Avoid descending of CS_fixer into non-code folders
   [#621](https://github.com/nextcloud/cookbook/pull/621) @christianlupus
+- Fixed compatiblity with Nextcloud 21
+  [#605](https://github.com/nextcloud/cookbook/pull/605) @icewind1991
 
 ## Deprecated
 - Obsolete routes to old user interface, see `appinfo/routes.php`
--- a/lib/Db/DbTypesPolyfillHelper.php
+++ b/lib/Db/DbTypesPolyfillHelper.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace OCA\Cookbook\Db;
+
+use OCP\Util;
+
+class DbTypesPolyfillHelper {
+	/**
+	 * @var String
+	 */
+	private $int;
+	/**
+	 * @var String
+	 */
+	private $string;
+	
+	public function __construct(Util $util) {
+		switch ($util->getVersion()[0]) {
+			case 18:
+			case 19:
+			case 20:
+				$this->int = \Doctrine\DBAL\Types\Type::INTEGER;
+				$this->string = \Doctrine\DBAL\Types\Type::STRING;
+				break;
+				
+			default:
+				$this->int = \OCP\DB\Types::INTEGER;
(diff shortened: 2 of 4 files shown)


seyfeb:
Hi @icewind1991, thanks for taking the iniative. Some things that I noticed immediately from looking at your PR (I did not check this out yet)

- I don’t think you should commit your `composer.lock` file (@christianlupus should we add this and maybe also `build/` to the `.gitignore`?)
- You need to sign off your commit (see DCO)
- You need to update the `CHANGELOG.md` file in the root to reflect your changes with a reference to this PR
- Most of the CI tests are failing. Are your changes compatible with older NC versions (<21)?

Did you check <https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#upgrading-to-nextcloud-21> and references therein for what needs to be done for NC 21 compatibility?

icewind1991:
>     * I don’t think you should commit your `composer.lock` file (@christianlupus should we add this and maybe also `build/` to the `.gitignore`?)

Removed, If you don't want them committed they should be added to the ignore yes

>     * You need to sign off your commit (see DCO)

Done
 
>     * You need to update the `CHANGELOG.md` file in the root to reflect your changes with a reference to this PR

Done

>     * Most of the CI tests are failing. Are your changes compatible with older NC versions (<21)?

The new constants aren't compatible with <21 yes, switched to just using the string literars instead

> Did you check https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#upgrading-to-nextcloud-21 and references therein for what needs to be done for NC 21 compatibility?

It's listed under the "Updated core libraries"

seyfeb:
I’m not sure about hardcoding the types. What about definining them version-dependent?

Maybe using the [NC version](https://nextcloud-server.netlify.app/classes/ocp-util#method_getVersion) in the constructor to assign them to a local variable (either `\Doctrine\DBAL\Types\Type::*` or `\OCP\DB\Types::*`)?

If I understand correctly, the [migration guide](https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/upgrade-guide.html#updated-core-libraries) suggests dropping them completely (temporarily) for multi-NC-version compatibility, but I must admit I don’t like this.

christianlupus:
I do not like to remove the type hints either. They are there for a reason. I asked [this question](https://help.nextcloud.com/t/keeping-development-in-sync-with-different-server-versions/104732) on the forum some time ago but only got an unofficial answer.
I suggest building a helper/polyfill class that returns the constants from the different classes depending on the NC version in use. Once NC 20 is EOL we can drop that class again. Does this sound reasonable in general, @icewind1991 @seyfeb?

One more thing I saw during a quick glance at the changes: You committed a whole bunch of whitespace changes (dropped indentation of otherwise empty lines). I know that some editors are doing this automatically. It makes the diff on the other side hard to read/check and cumbersome. This typically comes from a `git add -a` or committing via the graphical editor. It might be beneficial to keep the changes minimal.

christianlupus:
Closing this PR as it is already merged into branch `compatibility/nc21`.

Issue #604, opened 2021-02-22
Add Dashboard widget


Thovi98:
Hello ! First of all, I want to thank you all for this amazing app which have had a lot of improvements recently.

I’ve just had an idea I want to share with you : did you already think about a dashboard widget for this app ? I’ve thought about something like "Recipe of the day" which would randomly propose one recipe each day. 

Any opinion about this ? 

Of course, it’s not an urgent feature. 

Thank you again, it’s one of my favorite apps here !!

trvrcr:
Great idea, I would add the following:

I like my recipes being seasonal. Maybe add a field with te best period for the recipe and have the widget propose 1 at random from that season. 

I can open a new issue if you prefer having this separate.

seyfeb:
I think we should differentiate between data that is important and a core part for the recipe and data that is only handling displaying the recipe (here for a widget). 

E.g., a season for a recipe is best suited to be a keyword and fits well with the already existing structure. At which time a recipe containing a certain keyword is shown in a dashboard widget should not be part of the recipe data model but rather a setting of the dashboard widget.

I think it’s okay to use this issue to collect some ideas of what a Cookbook dashboard widget could look like.

Gtt1229:
I know this is older, but it's something I'm interesting it.

From my perspective, I would believe the widget should be random recipes, since "recent", "most used", etc isn't tracked. I also don't believe favorites make sense because from my understanding, widgets can't scroll.

I think it would be best used to serve random recipes at a glance so you can look at it and see if you would like to cook any of the "recommendations" at a glance.

I really just picture it with recipe image, name, and total time.

This could be expanded upon where breakfast, lunch, and dinner(Main Course) categories are dynamically "recommended" based on time of day.

Gtt1229:
I do not have PHP background, so what I have may require heavy criticism because I am heavily relying on Claude for this......

Here is the Widget as I am currently working on it (it is currently "snack time"):

<img width="312" height="501" alt="Image" src="https://github.com/user-attachments/assets/a7b6929e-e7ef-48ca-a9eb-116849562a6c" />

It implements time-based recipe recommendations in the following way:

```php
	private function bucketForHour(int $h): string {
		if ($h >= 5 && $h < 11) return 'breakfast';
		if ($h >= 11 && $h < 15) return 'lunch';
		if ($h >= 15 && $h < 17) return 'snack';
		if ($h >= 17 && $h < 22) return 'dinner';
		return 'snack';
	}
```
[...]

PR #611, opened 2021-02-24
Implement unified search provider

### LINKED ISSUE #559: Implement Search Provider for NC 20 unified search
Through the issues #479 and #474 I discovered the somewhat hidden search functionality for recipes. However, from other NC apps, I am used to using the unified search in the upper navigation bar. I would love for Cookbook to support this search functionality too. Searching by title would be enough for me, but of course this could be extended to ingredients or keywords.

Nextcloud 20 introduced the unified search. Every app can provide a search provider for its own content, with every matching entry being listed with a title, description and thumbnail. See details [here](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/search.html).

This functionality would be more intuitive and discoverable than the current search function. Even Ctrl+F or searching while in another app would work.

PFischbeck:
This fixes #559 by implementing a search provider for a unified search. Matching recpies are listed with their thumbnail (if available) and title, as well as the category if applicable.

search-image

The subline listing the category could easily be replaced by the recipe description, I am open for suggestions here. But I want to note that the current category lookup is expensive, as we read it from the recipe JSON file. Is that okay, or should we handle it differently?

### DIFF
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace OCA\Cookbook\Search;
+
+use OCA\Cookbook\AppInfo\Application;
+use OCA\Cookbook\Db\RecipeDb;
+use OCA\Cookbook\Service\RecipeService;
+use OCP\IL10N;
+use OCP\IUser;
+use OCP\IUrlGenerator;
+use OCP\Search\IProvider;
+use OCP\Search\ISearchQuery;
+use OCP\Search\SearchResult;
+use OCP\Search\SearchResultEntry;
+
+class Provider implements IProvider {
+
+	/** @var IL10N */
+	private $l;
+
+	/** @var IUrlGenerator */
+	private $urlGenerator;
+
+	/** @var RecipeDb */
+	private $recipeDb;
+
+	/** @var RecipeService */
+	private $recipeService;
+
+	public function __construct(IL10n $il10n, IUrlGenerator $urlGenerator, RecipeDb $recipeDb, RecipeService $recipeService) {
+		$this->l = $il10n;
+		$this->urlGenerator = $urlGenerator;
+		$this->recipeDb = $recipeDb;
+		$this->recipeService = $recipeService;
+	}
+
+	public function getId(): string {
+		return Application::APP_ID;
+	}
+
+	public function getName(): string {
+		return $this->l->t('Recipes');
+	}
+
+	public function getOrder(string $route, array $routeParameters): int {
+		if (strpos($route, 'files' . '.') === 0) {
+			return 25;
+		} elseif (strpos($route, Application::APP_ID . '.') === 0) {
+			return -1;
+		}
+		return 4;
+	}
+
+	public function search(IUser $user, ISearchQuery $query): SearchResult {
+		$recipes = $this->recipeService->findRecipesInSearchIndex($query->getTerm());
+		$result = array_map(
+			function (array $recipe) use ($user) : SearchResultEntry {
+				$id = $recipe['recipe_id'];
+
+				$subline = '';
+				$category = $this->recipeDb->getCategoryOfRecipe($id, $user->getUID());
+				if ($category !== null) {
+					// TRANSLATORS Will be shown in search results, listing the recipe category, e.g., 'in Salads'
+					$subline = $this->l->t('in %s', [$category]);
+				}
+
+				return new SearchResultEntry(
+					// Thumb image
+					$this->urlGenerator->linkToRoute('cookbook.recipe.image', ['id' => $id, 'size' => 'thumb']),
+					// Name as title
+					$recipe['name'],
+					// Category as subline
+					$subline,
+					// Link to Vue route of recipe
+					$this->urlGenerator->linkToRouteAbsolute('cookbook.main.index') . '#/recipe/' . $id
+				);
+			}, $recipes
+		);
+
+		return SearchResult::complete(
+			$this->getName(),
+			$result
+		);
+	}
+}
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -87,4 +87,7 @@ if (Util::getVersion()[0] >= 20) {
 			);
 		}
 	}
+} else {
+	class Provider {
+	}
 }
--- a/lib/Search/Provider.php
+++ b/lib/Search/Provider.php
@@ -12,74 +12,79 @@ use OCP\Search\IProvider;
 use OCP\Search\ISearchQuery;
 use OCP\Search\SearchResult;
 use OCP\Search\SearchResultEntry;
+use OCP\Util;
 
-class Provider implements IProvider {
+// IProvider requies NC >= 20
+// Remove conditional once we end support for NC 19
+if (Util::getVersion()[0] >= 20) {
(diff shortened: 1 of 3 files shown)


PFischbeck:
It fails for Nextcloud 18 and 19, because the interface for unified search was only added in Nextcloud 20. I see no easy way of supporting both pre-20 and 20 in the code.
NC 18 could be dropped, because it is EOL, but would dropping NC 19 be too drastic? We will probably run into the same problem if we tackle #604.

seyfeb:
I think the category is a good choice for the subline, as descriptions are typically too long to have be meaningful if only the first few words are shown. 

Regarding the category access: Maybe we could add a function for this to the Db service. Categories are stored in the database, so it should improve the performance.

Regarding older NC support: I would not drop NC 19 support just to have this new feature. However, can’t you check the NC version before registering the service? I’m not to firm in PHP but since it’s not compiled, I guess this should work(?)

christianlupus:
> I think the category is a good choice for the subline, as descriptions are typically too long be meaningful if only the first few words are shown.

Yes, I think this might be best suited at least fo now.

> Regarding the category access: Maybe we could add a function for this to the Db service. Categories are stored in the database, so it should improve the performance. @christianlupus certainly has an opinion on this ;)

As I am considering adding a DB abstraction layer that should handle these types of access quickly, you can use [this method](https://github.com/nextcloud/cookbook/blob/cc3082fa1dd22416dde7a1948210720dd2ee0430/lib/Db/RecipeDb.php#L498-L517) best. It returns the category from a recipe using only the DB. No need to manually traversing the file tree.

> Regarding older NC support: I would not drop NC 19 support just to have this new feature. However, can’t you check the NC version before registering the service? I’m not to firm in PHP but since it’s not compiled, I guess this should work(?)

We had recently a few issues regarding older NC instances (17 (!) and 18 if I remember correctly). So it seems there is quite a user base out there that is not living *on the edge*. I would rather avoid dropping support for officially supported NC versions (aka 19) yet.
There is a similar issue in #605 that we need to keep compatible with different NC versions. I think, here it might be useful to check [the NC version](https://nextcloud-server.netlify.app/classes/ocp-util#method_getVersion) and register only in NC >=20.

PFischbeck:
Thanks for your suggestions! I now use the method you suggested to retrieve the category from the DB.

I had also thought about checking the NC version before registering the search provider. However, the problem already lies in the definition of the `Application` class, which implements `IBootstrap`. If the `Application` does not implement this interface, the `register` function will not be called (see documentation [here](https://docs.nextcloud.com/server/latest/developer_manual/app_development/bootstrap.html)). However, this interface does not exist in NC < 20. Is there a way to poly-fill entire interfaces (`IBootstrap`, `IBootContext`, `IRegistrationContext`)?

Otherwise it seems like we will have to wait until NC 19 is EOL in June 2021... :disappointed:

christianlupus:
You could try to do something like the following (not sure if it really helps, but maybe worth a try):
```php
<?php
// This is the \OCA\cookbook\AppInfo\Application file
if (/* check if NC < 20 */) {
  namespace OCP\AppFramework\Bootstrap;
  interface IBootstrap {
    /* Specify the interface methods or even nothing just to make the name of the interface known */
  }
}

namespace \OCA\Cookbook\AppInfo;
use OCP\AppFramework\Bootstrap\IBootstrap;
class Application implements IBootstrap { /*...*/ }
```

PFischbeck:
Thanks for the suggestion! I have solved it in a slightly different way now, since I couldn't get the conditional definition of the interface in another namespace to work. Instead, I only conditionally define the `Application` and `Provider` classes. So this now runs in both NC19 and NC20+!
Should I add a changelog entry myself, or will you do that?

seyfeb:
You’re welcome to add the Changelog entry yourself :)

christianlupus:
One more thing I am seeing at the moment:
On NC19 the files are in face *empty*. no class is defined at all. I do not know if this is good as it could cause trouble if the autoloader scans a file and afterward cannot find the matching class (might be implementation-dependant). I suggest to have an else branch with an otherwise empty class definition.

PFischbeck:
I've made the suggested changes, although the `Application` definition could not be completely empty.

However, the PHP code style checker failure seems unrelated to my changes. I already have some findings about it: The recent minor update `2.18.4` of `php-cs-fixer` seems to be not backward-compatible, and we are not the only people hit by this, see [this issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5566). We'll have to wait for a new release, but such problems could be prevented by also committing the `composer.lock` file to the cookbook repository to pin the dependency versions, as recommended in the [Composer documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control). Is there any particular reason why we don't do it this way right now?

lib/Search/Provider.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace OCA\Cookbook\Search;
+
+use OCA\Cookbook\AppInfo\Application;
+use OCA\Cookbook\Db\RecipeDb;
+use OCA\Cookbook\Service\RecipeService;
+use OCP\IL10N;
+use OCP\IUser;
+use OCP\IUrlGenerator;
```
christianlupus, line 10:
```suggestion
use OCP\IURLGenerator;
```

lib/Search/Provider.php
```diff
@@ -0,0 +1,93 @@
...
+
+// IProvider requies NC >= 20
+// Remove conditional once we end support for NC 19
+if (Util::getVersion()[0] >= 20) {
+	class Provider implements IProvider {
+
+		/** @var IL10N */
+		private $l;
+
+		/** @var IUrlGenerator */
```
christianlupus, line 25:
```suggestion
		/** @var IURLGenerator */
```

lib/Search/Provider.php
```diff
@@ -0,0 +1,93 @@
...
+		private $urlGenerator;
+
+		/** @var RecipeDb */
+		private $recipeDb;
+
+		/** @var RecipeService */
+		private $recipeService;
+
+		public function __construct(IL10n $il10n, IUrlGenerator $urlGenerator, RecipeDb $recipeDb, RecipeService $recipeService) {
```
christianlupus, line 34:
```suggestion
		public function __construct(IL10n $il10n, IURLGenerator $urlGenerator,
			RecipeDb $recipeDb, RecipeService $recipeService) {
```

christianlupus (changes requested):
I was **not** able to search using this approach. Instead, I got errors in the nextcloud log regarding undefined return values. I stepped through the code and it seems the problem is with the case or `IURLGenerator`. With the changes mentioned here, I was able to search successfully. 

Please revisit these changes and check if you really had the correct case in your dev. Also what version were you developing against? I used a recent NC21 (hopefully, there is no undocumented change) under linux.

PFischbeck:
You're totally right about `IURLGenerator`, I applied the suggested changes. I'm running on NC21 via your `nextcloud-docker-debug` tool, and I'm pretty sure it did work, and I never changed the casing before. Weird. Either way, it should work now.

christianlupus (approved):
Seems good now to me. I just checked on my test install. Will merge if no issues arise during checks.

PR #623, opened 2021-02-27
Allow ordering recipe list

### LINKED ISSUE #594: Sort recipes by timestamp (created, edited)

Sometimes I want so see the last edited or created recipes first in the list; especially in the index but probably also in the categories (once there are lots of recipes assigned to a single category). 

Return timestamps with recipes and implement sorting recipes by date (created, modified) in index, categories, search (keyword & general). This would also allow showing the information on the recipe card. Add menu to select the order of the displayed recipes 

- alphabetically ascending
- alphabetically descending
- created ascending
- created descending
- edited ascending
- edited descending

Add ordering criteria to the server request, sort the list on the server and return sorted list.

seyfeb:
Closes #594

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## [Unreleased]
 
+### Added
+- Sorting recipes in list by creation and modification date   
+  [#623](https://github.com/nextcloud/cookbook/pull/623) @seyfeb
+
 ### Fixed
 - Minor errors in displaying ingredients and instructions
   [#642](https://github.com/nextcloud/cookbook/pull/642) @seyfeb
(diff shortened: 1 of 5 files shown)


christianlupus:
I just tested the PR. I had the following enhancements/suggestions (I assume rather quickly solvable):

- When the creation/modification times are shown, add a small icon similar to the one in the RecipeView. Just to be consistent.
- I had to try a bit to *unset* the sorting rule. Maybe it would be a good (?) idea to have filtering by name (both ascending and descending).
- Do we want to show the dates only when filtering is enabled? I get the idea but I suspect it might be feasible to show these always (it is only a small line of text).

Apart from these minor suggestions, I did not find any blocking issues. So if you think these changes should not be done directly, you can as well merge @seyfeb.

Issue #392, opened 2020-11-18
Improve public sharing of Recipes


basebuilder:
**Is your feature request related to a problem?**

As a NextCloud user, I like being able to publicly share rendered markdown notes from the Notes app like this:

- [Nicely rendered HTML](https://cloud.eotl.supply/s/6dXx9eLnkwcdP9t)

AFAICT as a Cookbook user, when sharing a recipe publicly, there is only the following:

- [File Directory](https://cloud.eotl.supply/s/GbHD2MYiWop43ki)
- [Download ZIP of Directory](https://cloud.eotl.supply/s/GbHD2MYiWop43ki/download)
- [JSON in HTML page](https://cloud.eotl.supply/s/GbHD2MYiWop43ki?dir=undefined&openfile=42003)

**Describe the solution you'd like**

Ideally, public sharing of recipes would by default display the same rendering currently displayed to an authenticated user within the Cookbook app- e.g.

![cookbook-recipe-view](https://user-images.githubusercontent.com/74658385/99549532-0b967e00-29ba-11eb-8897-4512618d6031.png)

Also,it would be helpful to display a pure JSON result by requesting something like

https://cloud.eotl.supply/s/GbHD2MYiWop43ki/json

**Describe alternatives you've considered**

Unless I completely overlooking something in the app, I don't know of an alternative solution at the moment.

christianlupus:
This is closely related #120 and also to #340.

You are right, currently there is no sharing possible (neither internally nor externally). There are some general changes to be addressed soon. The sharing will be one of these things.

basebuilder:
Thanks for clarifying @christianlupus I do PHP and front-end dev regularly, but have never worked with NextCloud apps.

Could you or someone else guesstimate how hard of task is in terms of learning your away around the APIs?

christianlupus:
Well it is not too hard to enter things. We define the web API ourselves. The nextcloud API is not too complex either. Currently some general preparations should be done in the codebase to help further development.

basebuilder:
I am new to the NextCloud App ecosystem especially Apps in the official `/nextcloud/app-name` organization. The Cookbook README says *"We're limited on resources"* does this mean in terms developers with time- is this constrained by money? 
Would it be helpful to help fund a dev to work on this feature (and maybe others)?

christianlupus:
Although not related with this issue, here are some points:

- The maintainer @mrzapp is searching for a person to replace him.
[...]

Issue #120, opened 2019-12-16
[Feature Request] Share recipe with other nextcloud users


GambaJo:
I would like to share my recipe for example with my wife (she has an account in our nextcloud). We don't want to have all the recipe twice.

mrzapp:
@GambaJo for the time being, you can just share the folder in your recipe library with, ask her to move it to her library and then either wait a minute or rescan the library.

Either that, or just have all your recipes in a shared folder

GambaJo:
> 
> 
> @GambaJo for the time being, you can just share the folder in your recipe library with, ask her to move it to her library and then either wait a minute or rescan the library.
> 
> Either that, or just have all your recipes in a shared folder

Ok, it's a little bit "tricky" but it works. Thx.

spikusmax:
HI there,

I second this request.  I have people I would like to share some/all my recipes with who are not highly skilled in the tech department but very active in the cooking - my mother mostly.  We are both avid cooks, but mum and I have very different cooking styles and values (I have scales that go to .01g, but she does not own a set!).  So the features that I'd be keen to see would be some way for  mum to be able to see my recipes, be able add her own to her own library and for me not to see her recipes (this is the issue with the shared recipe folder) unless she deliberately shares one with me, and then perhaps then I could 'Add'/copy to my library (a duplication as I may wish to modify the recipe whilst preserving her version).

Cheers,

Spike

mrzapp:
@spikusmax I hadn't though about it that way before. Maybe this whole sharing business is more of a copying business.

The main bulk of this work lies in the UX. When a user shares a recipe with another user, what happens? Does the other user just receive the recipe right away, or do they have to confirm it somehow? And how would this work on federated accounts?

The above need to be answered before we move on with implementation.

drewgreat:
I'd like to be able to share to non users as well.  Either a single recipe or the entire thing (ideally, by tag or something too, but don't need to get too carried away).  It should work the same way gallery share links do (by opening the image up).  I love the app, but I was surprised to see that it was allocated to a single directory I kinda thought it would work more like the gallery app does and could be opened on any folder that contains recipes.

mrzapp:
@drewgreat the general consensus is to actually move away from files towards database only storage in the future, as it makes indexing for search a lot easier.

I think we need that improvement inplemented before we start working on the sharing issue, as there are probably standard ways to do this in NextCloud, which we'd want to follow

drewgreat:
[...]

PR #627, opened 2021-03-02
Sharing Options

matrois:
Sharing options would be handy. At its best one for recipe and one for category (i.e. to allow public access on whole categories)


seyfeb:
You probably meant to open an issue instead of a pull request. There are already two issues discussing sharing, see #120 #392

Issue #626, opened 2021-03-02
500 error with empty alert with link from specific website


littlepiggie:
**Description**
When I insert a particular link, the server errors and I get an alert without message.

**Reproduction**
1. Enter in "Download from URL" `https://www.ah.nl/allerhande/recept/R-R1185300/maanzaadbolletjes`.
2. Click button.
3. Empty popup, 500 internal server error in console.

**Expected behavior**
No 500 error. Clean message saying website is unsupported.

**Browser**
Firefox

**Versions**
Nextcloud server version: 20.0.1
Cookbook version: 0.8.1

littlepiggie:
Note: the website uses the schema format, so it should work.

cloud2018:
Perhaps the cookie banner on the website prevents the recipe from downloading correctly

christianlupus:
The reason for this issue is that the website uses the context `http://www.schema.org/Recipe` in their microtypes. See here in line 308:
```
<article itemscope itemtype="http://www.schema.org/Recipe">
```
Our parser only looks for the `http://schema.org/Recipe` microtype context. Therefore the recipe is not found.
When implementing #553 this issue can be tackled as well.

christianlupus:
At least now, there should be a better error message present.

cvinhaes:
Hi there. Same here with XHR POST https://drive.cloud-io.network:2053/apps/cookbook/import [HTTP/2 500 Internal Server Error] and empty error message box in Firefox 87.0. I'm running Nextcloud 20.0.0. Same issue happens with many brazilian recipe sites, one example:
https://receitas.ig.com.br/guiadacozinha/2021-04-11/surpreenda-a-familia-com-a-receita-de-costelinha-de-porco-com-chutney-de-abacaxi.html

christianlupus:
> https://receitas.ig.com.br/guiadacozinha/2021-04-11/surpreenda-a-familia-com-a-receita-de-costelinha-de-porco-com-chutney-de-abacaxi.html

This page does not support schema.org recipe meta data. Therefore, no import is possible ATM.

cgudea:
It appears the recipes from SeriousEats have the same issue even though the scheme.org recipe metadata exists. Serious eats did work before

Example recipes:

* https://www.seriouseats.com/recipes/2015/02/grilled-lemongrass-coriander-marinated-tofu-vietnamese-sandwich-vegan-recipe.html
[...]

PR #647, opened 2021-03-06
Show error message upon import exception

christianlupus:
This solves partially the issue described in #626 as it avoids an empty alert box.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
   [#653](https://github.com/nextcloud/cookbook/pull/653) @seyfeb
 - Setting nutrition information on recipes with an array assigned for nutrition
   [#653](https://github.com/nextcloud/cookbook/pull/653) @seyfeb
+- Fix empty error message upon import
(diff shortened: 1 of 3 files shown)


christianlupus:
@seyfeb as the original line in the `AppNavi.vue` came from the Axios update,  I just wanted to make sure you are ok with this modification. Or should we check that at least some result was returned before we try to use it.

seyfeb:
> @seyfeb as the original line in the `AppNavi.vue` came from the Axios update, I just wanted to make sure you are ok with this modification. Or should we check that at least some result was returned before we try to use it?

I think there is always some response from the server? To be sure one could provide a default feedback to the user if it’s empty.

christianlupus:
@seyfeb I had a look at the axios documentation once more and came up with a bit more *secure* code in the sense that exceptions are handled more explicitly.

seyfeb:
I think this should be fine, except for the failing linters 😄

christianlupus:
The linters are taken care of.
Good to have them in the CI chain :smile:

PR #668, opened 2021-03-27
Updated styling of code according to cs fixer

christianlupus:
This should cover the issues regarding the new code styling issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   [#653](https://github.com/nextcloud/cookbook/pull/653) @seyfeb
 - Fix empty error message upon import
   [#647](https://github.com/nextcloud/cookbook/pull/647) @christianlupus
+- Update code styling to match with current version of php-cs-fixer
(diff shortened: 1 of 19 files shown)


PFischbeck:
Just as heads-up: the next version of PHP CS Fixer will apparently reverse the enforced behavior again, since this was an implementation error on their end. See [this issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5566) and [this PR](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5567).

christianlupus:
> Just as heads-up: the next version of PHP CS Fixer will apparently reverse the enforced behavior again, since this was an implementation error on their end. See [this issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5566) and [this PR](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5567).

Thanks for the hint. We will have to deal with it then. But it should not be too critical. Everything was automatically fixed.

PR #678, opened 2021-04-06
Enhanced tag cloud

### LINKED ISSUE #643: Ability to hide tag cloud
The tag cloud at the top of the main page takes up screen real estate and cannot be hidden. At the same time, the space isn't large enough to be actually useful. At the moment, it's just in the way.

A toggle, checkbox, or radio button which would allow users to display or hide recipe tags

Moan about it in the issues section

### LINKED ISSUE #660: Sort keywords alphabetic
I'm always frustrated when I want to search a specific recipe and the tags/keywords of a recipe or for all recipes are sorted by creation date, not alphabetic.

Sort keywords in RecipeList, RecipeView and RecipeEdit always alphabetic, which makes it faster to search for specific or probably similar keywords and allows user to sort them their selves by e.g. using "1.", "2.", …

Probably sort by anything else, but it should be consistent, ~but also not depend on creation date like an incrementing database id, as this does not allow users to determine the order other than by deleting and re-creating all keywords.~ (after realizing they are currently sorted by number), but please do not sort them by count of recipies or let the user decide about the order, because IMO sorting them by count makes the sorting not really helpful, but painful because after adding/editing a recipe, you might need search again for a special tag because it changed its place.

seyfeb:
- [x] Setting for hiding keywords (local browser setting), closes #643 
- [x] Resizing tag cloud
- [x] Order keywords either by name or count, closes #660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### Added
 - Make recipes searchable through unified search
   [#611](https://github.com/nextcloud/cookbook/pull/611) @PFischbeck
+- Enhanced keyword cloud in recipe list with option to hide/show keywords, enlarge area, and ordering alphabetically
+  [#678](https://github.com/nextcloud/cookbook/pull/678) @seyfeb
 
 ### Fixed
 - Calling reindex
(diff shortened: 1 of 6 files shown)


christianlupus (approved):
Seems good to me and function is given in a quick test.

PR #702, opened 2021-04-28
Generate clear message when recipe exists during creation

### LINKED ISSUE #690: Show correct notification when recipe already exist
When I add a recipe which already exists, I just get the browser notification "Recipe could not be saved".
This isn't a big problem if you just have several recipes, because you see them all on the screen. But if you host hundreds of them, this message is a bit confusing.

Change the browser notification to a proper one.

Ask if the user want's to replace the existing recipe with the current or force to change the name of the curren recipe.

christianlupus:
Closes #690

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
   [#699](https://github.com/nextcloud/cookbook/pull/699) @christianlupus
 - Enable stalebot
   [#700](https://github.com/nextcloud/cookbook/pull/700) @christianlupus
+- Correct error messages when recipe already exists
+  [#702](https://github.com/nextcloud/cookbook/pull/702) @christianlupus
   
 ## 0.8.4 - 2021-03-08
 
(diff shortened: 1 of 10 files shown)


christianlupus:
@seyfeb I am getting some strange errors. It seems I have tomatoes on my eyes. Can you have a quick glance if you see the culprit?

I get this error: 
grafik
to corelate the line numbers, here the debugger's output
grafik

Can you explain why the exception is passed on to the browser although it is caught?

seyfeb:
I pulled the PR and can't reproduce this behavior. I get the XHR error but no other. Maybe you need to clear your cache?

christianlupus:
The issue is merely esthetical. The main functionality is given as far as I see.

In a new browser (private window), the issue persisted. Also rebuilding using `npm run build` did not solve it. What else could it be? The node.js version (14 for me)?

Eventually, I could try to use the zip file from the GH actions to see if this is something with my machine/build setup.

PR #718, opened 2021-05-12
Half the work done in github actions

christianlupus:
Currently, all tests are run twice, once in the pull request and once for the push itself. This should fix this and speed up PR tweaking a bit.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,14 +1,13 @@
 
-name: CI-tests
+name: Continuous integration testing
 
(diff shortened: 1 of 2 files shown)


christianlupus:
This should **not** be merged, as depiced in [this link](https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-does-pull_request-affect-actionscheckout). the `push` and `pull_request` serve different purposes and will test a different commit (merge commit vs latest branch head). So we should stick with both.

PR #719, opened 2021-05-12
Remove unneeded dependency

christianlupus:
This will remove an unneeded NPM dependency as far as I can see.

This dependency seems to be a leftover of the original implementation work and (currently) is references nowhere in the codebase.


seyfeb (approved):
I don’t know of any usage of the dependency either

PR #747, opened 2021-06-18
Updated/added screenshots for the Nextcloud App Store

### LINKED ISSUE #635: Update image in NC app store
Add an up-to-date image with a better resolution of the Cookbook app for the Nextcloud app store. (I think it’s the image in `img/screenshot-small.jpg`). The current one looks really blurry [on the NC homepage](https://apps.nextcloud.com/apps/cookbook).

mMuck:
1920p width, 2 different ones.
Bug fixed: The **appinfo/info.xml** referred only to the thumbnail as the full size screenshot.
Resolves #635 - this is my first pull request. Surly I did something wrong with my request...


seyfeb:
@mMuck when you click on 'Details' of the failed workflow (in your case DCO), it shows you instructions on how to fix this PR! :)

Screenshot 2021-06-21 at 10 54 25

christianlupus (changes requested):
Please change the template in `.github/actions/deploy/info.xml.dist` as well as indicated in line 2. Otherwise, the changes to the `info.xml` will be lost upon next release cycle.

A side remark: The parser on the app store requests the XML file elements to be in a defined order. If the order is wrong, the elements are ignored. Please check this (I did not do it).

Issue #764, opened 2021-07-07
Nextcloud 22 compatibility


ben16w:
Hi, Nextcloud 22 was released today. Will Cookbook be updated to support it?

Many thanks to everyone who's worked on this project. I use it almost daily!

christianlupus:
It will be made available for NC 22. Currently, we have to settle #763/#762 before a new release can be published.

ben16w:
Ok thanks for the info @christianlupus

MoralCode:
after looking at the two issues mentioned, it seems as though both are resolved, nearly resolved, or only applicable to some usecases (like shared instances).

Is there a current summary of which users should and should not attempt to update the nextcloud cookbook app?

From my brief reading of the issues it seems like if you 1) do not share your instance with anyone and 2) back up your cookbook folder, delete the app, and install version 0.9.1 manually/fresh it should be safe to perform this upgrade, even though this version is not listed on the store due to the mentioned issues making it an upgrade unsuitable as a matter of general advice for all users.

Is this a correct analysis of the problem? do others who are more familiar with this issue agree that the described set of users are in the clear as far as upgrading?

christianlupus:
If there are no further (yet hidden) problems in the new version, it seems pretty well explained. I am preparing a fix for all cases I can think of but yet the PostgreSQL code is missing/faulty and the automated tests cause trouble at the moment.

The way removing the app installing the 0.9.1 version from GitHub directly should (unless we missed another bug yet) work. You will lose the cached versions of your recipes so the first start of the app might take quite some time for each user using the app. If you had a really big cookbook folder, you might run into timeouts by your web server. You could test this by creating a new NC user and copying the recipes over. The parsing was not modified so when this test runs smoothly, you should not have trouble.

PR #778, opened 2021-08-09
Mark nextcloud veriosn 22 compatibility

christianlupus:
This will mark the app as compatible with NC 22.

Some enhancements in the testing/deployment procedure are added as well.


christianlupus:
Closes #764

Issue #837, opened 2021-10-29
Update npm before building


christianlupus:
In the actions the npm command is not globally updated prior to the main work. This should be changed.

PR #839, opened 2021-10-30
Some repair steps for the github actions

christianlupus:
This should remove some of the warnings at the npm building during the GH action runs.


christianlupus:
Closes #837

PR #886, opened 2022-02-08
Filter empty pasted lines before single line check

MarcelRobitaille:
I noticed this while trying to implement something else.

When pasting multiple lines, the empty lines should be filtered out before checking if only one line was pasted, shouldn't they?

For example, if the following is pasted: `"this is a test\n"`, this will be split into `["this is a test", ""]` and `this.singleLinePasted` will be false even though later the empty item will be filtered out and the array will contain a single element.

I also replaced the for loop with the filter method, I hope that's ok. I think it's more clear what is going on.


seyfeb (approved):
I didn’t check and run this, but it look good to me. I also prefer your solution of using a filter method - makes sense!

Issue #834, opened 2021-10-28
Always show save button in the recipe editor


seyfeb:
**Problem** 

In the recipe-editing window the save button is shown in the top bar or at the very end of the edit fields. When editing something in the middle on small screens one needs to scoll to get access to the button.

**Possible solutions**

1. Show the bottom button as a floating button in the bottom right corner while the top-bar button is not visible. Beware (and test) if this also works on small screens (i.e., smartphones).
2. Fix the top bar such that it is always visible when scrolling.

PR #888, opened 2022-02-09
Make AppControls breadcrums sticky (#834)

MarcelRobitaille:
Make the buttons always visible so that they are accessible even when
scrolled halfway down a recipe.

This is consistent with the "Open navigation" / "Close navigation" button and gives more information and controls than making the bottom save button sticky.

image

### DIFF
--- a/src/components/AppControls.vue
+++ b/src/components/AppControls.vue
@@ -346,6 +346,15 @@ export default {
 .wrapper {
     width: 100%;
     padding-left: 4px;
+    /* Sticky is better than fixed because fixed takes the element out of flow,
+     which breaks the height, putting elements underneath */
+    position: sticky;
+    /* The height of the nextcloud header */
+    top: 50px;
+    /* This is competing with the recipe instructions which have z-index: 1 */
+    z-index: 2;
+    background-color: var(--color-main-background);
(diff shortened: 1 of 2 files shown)


src/components/AppControls.vue
```diff
@@ -344,8 +344,19 @@ export default {
...
 .wrapper {
+    /* Sticky is better than fixed because fixed takes the element out of flow,
+     which breaks the height, putting elements underneath */
+    position: sticky;
+
+    /* This is competing with the recipe instructions which have z-index: 1 */
+    z-index: 2;
+
+    /* The height of the nextcloud header */
+    top: 50px;
```
christianlupus, line 355: I honestly do not like to do it this way. Once, the styling of the server gets changed, we will run into trouble. But for now, let's keep it simple and hope for the best.

MarcelRobitaille: I agree. I dislike such magic numbers. I looked for some API that nextcloud core would expose, but I didn't find anything.

Edit: I just looked through the `:root` variables in dev tools and found `var(--header-height)` which is set to `50px`.

christianlupus: This is resolved in #897.

Issue #882, opened 2022-01-17
Switching to next instruction row impossible


christianlupus:
**Description**
It should be possible to switch to the next instruction row by pressing enter (see [this snippet](https://github.com/nextcloud/cookbook/blob/d5d41a31d4e7defe6d6f1b6d003749747f59c69b/src/components/EditInputGroup.vue#L247-L249)). However, an error is printed to the console.

**Reproduction**
1. Edit a recipe
2. Make sure it has multiple instruction rows.
3. Select any instruction row but the last one
4. Hit enter

**Expected behavior**
The next row is put to focus.

**Actual behavior**
Nothing happens visually. An error is logged to the console.

The reason is the fact that the instructions are `textarea` elements and not `input` elements. The search in [this line](https://github.com/nextcloud/cookbook/blob/d5d41a31d4e7defe6d6f1b6d003749747f59c69b/src/components/EditInputGroup.vue#L248) will cause no matching DOM element to be found.

MarcelRobitaille:
I am able to reproduce this. I would like to work on this issue.

PR #890, opened 2022-02-09
Enable enter to go to next instructions step

MarcelRobitaille:
This is a fix for #882.

When enter is pressed in `EditInputGroup`, there is an attempt to select the next input field. This works for ingredients, which use inputs, but not for instructions, which use `textareas`. Now, both are searched for using `querySelector`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
   [#897](https://github.com/nextcloud/cookbook/pull/897) @MarcelRobitaille
 - Fix UI glitch when keyword list is empty
   [#892](https://github.com/nextcloud/cookbook/pull/892) @MarcelRobitaille
+- Allow switching to new instruction line with Enter key
+  [#890](https://github.com/nextcloud/cookbook/pull/890) @MarcelRobitaille
 
 ### Documentation
 - Added clarification between categories and keywords for users
--- a/src/components/EditInputGroup.vue
+++ b/src/components/EditInputGroup.vue
@@ -287,8 +287,11 @@ export default {
                     ) {
                         this.addNewEntry()
                     } else {
+                        // Focus the next input or textarea
+                        // We have to check for both, as inputs are used for
+                        // ingredients and textareas are used for instructions
                         $ul.children[$pressedLiIndex + 1]
-                            .getElementsByTagName("input")[0]
+                            .querySelector("input, textarea")
                             .focus()
                     }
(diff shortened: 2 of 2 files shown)


seyfeb:
Actually this is not an oversight. What if you wanted to have a line break within your instruction? If we decide to jump to the next line by pressing enter for consistency, we should at least provide a possibility (e.g., shift+enter) to insert line breaks.

MarcelRobitaille:
Hi @seyfeb. Thanks for reviewing this. I thought of that as well. Maybe we could have one modifier key to insert a newline and another modifier to go to the previous textfield. Or maybe enter alone would be newline, ctrl+enter would be next textfield, and shift+enter would be previous textfield. What do you think?

seyfeb:
I’m not sure if I have seen such a thing anywhere, but is it a common UI pattern to have "Return/Enter" jump to the previous item?

Or are you suggesting that the (modifier+)enter action should insert a new empty field above the current field?

MarcelRobitaille:
I was suggesting to jump to the previous field, not create a new one. I guess you are right, I have only seen this with tab / shift+tab.

seyfeb:
Thinking about this I find the idea to have a possibility to insert a new field above or below the current field with shortcuts also interesting... This should however become a separate issue and be discussed there.

I see two lines of reasoning:

1. Keep the current behavior and use "Enter" to insert a line break in the current instruction text field because that is how it behaves currently. Use (shift/ctrl/alt)+enter to jump to the next instruction field.

2. Make the behavior consistent with the ingredients field ("Enter" jumps to next field) and use (shift/ctrl/alt)+enter  to create a newline

@christianlupus any thoughts on this?

christianlupus:
I would use the same UX as with the tools and ingredients: simple enter switches to next row or spend one. Shift plus enter adds a meeting into the text area. 
I think going up is not so critical. For one, there is the tab key. Further, i associate enter with going downward. One could even consider using the arrow keys to navigate over the top of a textbox to the bottom of the box above.

The inserting of intermediate steps is too be discussed in a separate issue in my opinion.

seyfeb:
Let’s go with that way then: Enter to go to the next item as for ingredients. For inserting a newline I would suggest shift+enter as I have seen this in multiple messengers, so it seems to be a common (and therefore somewhat expected) behavior

christianlupus:
I force-pushed this branch to fix the DCO check. @MarcelRobitaille you need to rebase your local development accordinlgy.

PR #891, opened 2022-02-09
Document development environment setup

MarcelRobitaille:
I had some issues figuring out how to set up my development environment, so I thought I would document it after figuring everything out.


christianlupus:
I have some comments to add here. O will do so or open another or once this is merged.

Issue #857, opened 2021-11-25
Empty list of keywords causes UI glitch


christianlupus:
**Description**
If editing a recipe without a keyword attached, there is a glitch in the UI.

**Reproduction**
1. Open the edit dialog of a recipe without keywords
2. Look at the keyword selection dialog

**Expected behavior**
See an empty box

**Actual behavior**
A bubble in the box is visible. This seems due to a bad design decision for the API communication.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/8202922/143435663-db83b990-d736-4fd6-ba03-9fc9d1715686.png)

**Browser**
Firefox + Chrome under linux

**Versions**
Cookbook version: 0.9.6 (master branch)

**Additional Information**
It might be favorable to update the API to always use an array instead of a packed string. That would solve the problem most probably.

PR #892, opened 2022-02-10
Fix the UI glitch when keywords are empty (#857)

MarcelRobitaille:
Fix the UI glitch with a tiny blob in the keywords field when loading a recipe with empty keywords. The bug was caused by `keywords.split(',')` when `keywords` is an empty string. This will create an array of a single element: an empty string. Rendering this empty string as a keyword results in this little blob. The solution is to filter out empty strings after the call to `.split()`.


seyfeb (approved):
Again, untested but looks good ;)

christianlupus:
closes #857

PR #900, opened 2022-02-28
Prevent newlines switching input fields

### LINKED ISSUE #881: Newlines are not prevented when switching input fields
**Description**
Currently, the [prevention of keyevents](https://github.com/nextcloud/cookbook/blob/d5d41a31d4e7defe6d6f1b6d003749747f59c69b/src/components/EditInputGroup.vue#L232) should avoid inserting a newline symbol if using enter to add a new entry. Unfortunately, the `keyup` event [does not allow this](https://stackoverflow.com/a/14055191/882756). As a result, the newline is added nontheless.

**Reproduction**
1. Edit a recipe
2. Add a new instruction row and/or go to the very last instruction text area
3. Press Enter

**Expected behavior**
A new area/step is added. The new step is empty. The old one is unaffected by the enter pressing.

**Actual behavior**
The newline is in the text area and later in the recipe visible.

**Screenshots**

**Browser**
Which browser are you using? Firefox

MarcelRobitaille:
Fixes #881.

Use a `keydown` event listener rather than a `keyup` event lister. This allows us to use `event.preventDefault()` when enter is pressed to go to the next textarea without adding a `\n` to the current textarea.

I believe the reason to use `keyup` previously was to not get repeat events. In order to make sure that repeat events are still not considered, we exit early if `event.repeat` is true.

I also took this opportunity to clean up some parts of the code.
First, I replaced `event.keyCode === 13` with `event.key === 'Enter'`. This was done for many reasons.
1. [`event.keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)
2. `event.key === 'Enter'` is more readable than `event.keyCode === 13`. You do not have to lookup what 13 means.
3. `event.keyCode === 10` was removed. I am not sure what this check comes from. I cannot find keycode 10 in any table.

The last change was to remove the nested if statements.
Instead of checking if the key is <kbd>enter</kbd> or <kbd>#</kbd> to prevent default
and then handling each individually,
I exit early if the key is neither.

Previous behaviour (when I hit enter with my cursor before `test`, the current input becomes `\ntest` before a new textarea is created):

https://user-images.githubusercontent.com/8503756/156021618-7c62da26-6a40-45ac-b5b3-30472eecdd84.mp4

New behaviour (hitting enter before `test` or inside of `test` does not add `\n`):

https://user-images.githubusercontent.com/8503756/156021637-b0542746-6d0b-4333-abcb-7ba69f1dc2d2.mp4

Repeat events are still ignored (I am holding the enter key):

https://user-images.githubusercontent.com/8503756/156021875-45ffc874-e2b1-4790-a3ec-dc0d70f1168f.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
   [#892](https://github.com/nextcloud/cookbook/pull/892) @MarcelRobitaille
 - Allow switching to new instruction line with Enter key
   [#890](https://github.com/nextcloud/cookbook/pull/890) @MarcelRobitaille
+- Prevent inserting newline characters in instructions/ingredients/tools when pressing enter
+  [#900](https://github.com/nextcloud/cookbook/pull/900) @MarcelRobitaille
 
 ### Documentation
 - Added clarification between categories and keywords for users
--- a/src/components/EditInputGroup.vue
+++ b/src/components/EditInputGroup.vue
@@ -271,54 +271,57 @@ export default {
                 return
             }
 
+            // Only do anything for enter or # keys
             if (
-                e.key === "Enter" ||
-                (this.referencePopupEnabled && e.key === "#")
+                e.key !== "Enter" &&
+                !(this.referencePopupEnabled && e.key === "#")
             ) {
-                e.preventDefault()
-                const $li = e.currentTarget.closest("li")
-                const $ul = $li.closest("ul")
-                const $pressedLiIndex = Array.prototype.indexOf.call(
-                    $ul.childNodes,
-                    $li
-                )
+                return
+            }
 
-                if (e.key === "Enter") {
-                    if (
-                        $pressedLiIndex >=
-                        this.$refs["list-field"].length - 1
-                    ) {
-                        this.addNewEntry()
-                    } else {
-                        // Focus the next input or textarea
-                        // We have to check for both, as inputs are used for
-                        // ingredients and textareas are used for instructions
-                        $ul.children[$pressedLiIndex + 1]
-                            .querySelector("input, textarea")
-                            .focus()
-                    }
-                } else if (this.referencePopupEnabled && e.key === "#") {
-                    e.preventDefault()
-                    const elm = this.$refs["list-field"][$pressedLiIndex]
-                    // Check if the letter before the hash
-                    const cursorPos = elm.selectionStart
-                    const content = elm.value
-                    const prevChar =
-                        cursorPos > 1 ? content.charAt(cursorPos - 2) : ""
+            // If it is one of those keys, prevent the browser's default action
+            e.preventDefault()
 
-                    if (
-                        cursorPos === 1 ||
-                        prevChar === " " ||
-                        prevChar === "\n" ||
-                        prevChar === "\r"
-                    ) {
-                        // Show dialog to select recipe
-                        this.$parent.$emit("showRecipeReferencesPopup", {
-                            context: this,
-                        })
-                        this.lastFocusedFieldIndex = $pressedLiIndex
-                        this.lastCursorPosition = cursorPos
-                    }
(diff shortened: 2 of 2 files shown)


seyfeb:
> event.keyCode === 10 was removed. I am not sure what this check comes from. I cannot find keycode 10 in any table.

Not sure, but this might have had sth to do with https://bugs.chromium.org/p/chromium/issues/detail?id=79407 I guess it is save to remove it.

Otherwise this PR looks fine.

christianlupus:
I can tell you, @MarcelRobitaille that the 10 and 13 are the ASCII codes of the carriage return (`\r`) and newline (`\n`) characters. On linux, only the `\n` (aka 13) is used as line termination, on Mac it is only `\r` and on Windows it is both of them (`\r\n`).

Most probably this with `10` is a leftover of some compatibility issue.

christianlupus (changes requested):
There is a regression in the code introduced: The referencing of recipes does no longer works. When pressing as `#` a popup should appear and allow the user to select a recipe to reference. This popup is not opening anymore.

MarcelRobitaille:
I see. I did not know about that behaviour. I will try to resolve the regression.

MarcelRobitaille:
I found the issue. There was an off-by-one error because we are using keydown instead of keyup now, so the new character did not exist in the textarea at the time that we were checking for the character before. The popup would actually work if you type two whitespace characters and then a pound.

christianlupus:
That sounds reasonable! I will give it once more a test.

MarcelRobitaille:
Thank you. I think the code is still understandable. I added a comment about how the keydown event fires before the new character is in the input. Alternatively, we could have a keydown event listener for enter and a keyup event listener for #. It would be more code, but it might make the flow more clear if each handler is only doing a specific job.

christianlupus:
Hello again. I fear the two individual event handlers might be a better solution. Now another issue has been triggered: Again we are in the case of `#`. The result should be `#r/1234`. In the description field, you can see this. In the `EditInputGroup` fields, the `preventDefault()` will prevent the printing of the initial `#` thus resulting only in `r/1234`. This will not be detected and replaced with ha link later on.

I thought whether it was favorable to prevent the `#` in all locations by `preventDefault` and adding it from the event there or keeping the `#` and adding the part `r/1234` from the event of the modal dialog. I think of the case if the user wants a literal ´#`. Then with `preventDefault` we would need some additional link as the `#` is never printed in the text field. Thus, I think, in the case of `#` the `preventDefault` must be delayed or the sharp must be added programmatically.

MarcelRobitaille:
Hi @christianlupus. I noticed that as well. I didn't know what the intended behaviour was. I only added `e.preventDefault()` because the previous code had it, but in the case of a keyup listener it wasn't doing anything. Is there any reason not to remove it? I think we only want to prevent the default action for enter. For the `#`, we want it to appear in the input and to trigger the popup, right? Alternatively, if you only want the `#` to be inserted after selecting something in the popup, we can prevent the default event then add the `#` programmatically. Just let me know and I can update the code.

MarcelRobitaille:
Please also let me know if I should split it into a keydown and a keyup listener. It would not be difficult.

christianlupus:
I think we are overstretching this PR with too much content. This was partially broken in the past, and might as well be related to #901. Thus, I consider it separated.

> Please also let me know if I should split it into a keydown and a keyup listener. It would not be difficult.

This I consider only an implementation detail. I have no clear favorite for either of them. I am just considering that mixing different event listeners brought us to #901. Thus, I *personally* prefer the `keydown` event. Any hints from your side, @seyfeb?
Anyways, I would say this is to be discussed separately.

I also opened #903 to discuss the desired behavior for the user before starting to implement the code.

PR #909, opened 2022-03-07
Make Vue hot reload of components possible

christianlupus:
This should make it possible to do hot component replacement as defined by Vue maintainers.


christianlupus:
There are some obsolete packages in the package.json as reminders. I will not remove them here as this will blow up the PR's size. INstead, a new issue will be opened accordingly.

PR #910, opened 2022-03-08
Revert Nextcloud core's contentedibtale styles

### LINKED ISSUE #908: Small input fields in recipe creation on mobile device
**Description**
On my mobile device the input fields, especially the description is far too small 

**Expected behavior**
It should fill the width of the window. for other fields like name probably as well 

**Screenshots**
Screenshot_20220305-213233_Firefox_1

**Browser**
FF 97.2.0 on android

**Versions**
Nextcloud server version: 21
Cookbook version: 0.9.10

MarcelRobitaille:
Fixes #908 

Nextcloud core is setting global unscoped styles on any `div[contentedibtale]`. Such a div is used by our markdown editor:
CodeMirror. Fix this by overwritting Nextcloud core's heavy-handed global styles.

Now, adding `contenteditable="true"` does not change the style inside our markdown editor:

https://user-images.githubusercontent.com/8503756/157314618-a553dee7-6857-4f63-a949-9eb793d93201.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### Fixed
 - Reduce complex coupling between event handlers in EditInputGroup.vue
   [#901](https://github.com/nextcloud/cookbook/pull/901) @MarcelRobitaille
+- Fix bug in NC Vue config that switches input fields to be not full width on mobile
+  [#910](https://github.com/nextcloud/cookbook/pull/910) @MarcelRobitaille
 
 ### Documentation
 - Introduction about how to start coding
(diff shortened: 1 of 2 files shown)


seyfeb:
Good catch! Just tested this directly on my phone and works as expected :)

@MarcelRobitaille What do you think about the width of the Name and URL fields? Should we make them 100% on mobile as well?

MarcelRobitaille:
@seyfeb thanks for checking the PR.

> What do you think about the width of the Name and URL fields? Should we make them 100% on mobile as well?

I think it would be consistent. I will try to add that too.

PR #914, opened 2022-03-12
Don't prevent default event of inserting # when linking recipes

### LINKED ISSUE #905: Fix the linking of recipes in tools/ingredients/instructions
**Description**
There is a bug in the implementation of the `EditInputGroup` that causes a missing `#` symbol during adding a link to another recipe.

See also [this comment](https://github.com/nextcloud/cookbook/pull/900#issuecomment-1059367049).

**Reproduction**
1. Edit/create a recipe
2. Add a tool, ingredient, or instruction
3. Type a sharp symbol
4. Select one recipe

**Expected behavior**
The text contains a correct link like `#r/1234`.

**Actual behavior**
The initial `#` is not printed.

**Versions**
Cookbook version: 0.9.10

MarcelRobitaille:
Fixes #905 

When refactoring they keydown handler of EditInputGroup, a regression was introduced where the # would not be inserted when linking to another recipe. This is because we previously had a keyup event listener, which cannot prevent the key from being insterted, but now we are using a keydown listener, which can. Therefore, the `e.preventDefault()` that was already present in the code now has unintended side effects. We can remove it, there is no reason to have it there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
   [#912](https://github.com/nextcloud/cookbook/pull/912) @christianlupus
 - Remove deprecation in preparation for Sass 2.0.0
   [#915](https://github.com/nextcloud/cookbook/pull/915) @MarcelRobitaille
+- Fix regression in #900 to allow inserting links to other recipes again
+  [#914](https://github.com/nextcloud/cookbook/pull/914) @MarcelRobitaille
 
 ### Documentation
 - Introduction about how to start coding
--- a/src/components/EditInputGroup.vue
+++ b/src/components/EditInputGroup.vue
@@ -313,7 +313,6 @@ export default {
                     cursorPos === 0 ||
                     /\s/.test(content.charAt(cursorPos - 1))
                 ) {
-                    e.preventDefault()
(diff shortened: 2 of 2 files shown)


christianlupus:
No, I cannot confirm this to be a fix. At least on my machine (Archlinux Firefox) this triggers a bit of a different behavior: When I type a `#` in on of the InputGroupFields, a popup opens. This list in the popup is _empty_. The reason is that the `#` is prefilled in the search field. Removing it allows to select and add a link. Still, however, the `#` is not inserted in the final result. So, I still get `r/1234` instead of `#r/1234`.

christianlupus:
I made a screencast to show my problems. This is different for you?

https://user-images.githubusercontent.com/8202922/158067642-79adfe34-5aab-4e88-8a77-1d5581c5089d.mp4

MarcelRobitaille:
That is very strange. I can reproduce this. It seems that the `contenteditable` markdown editor behaves differently to `input` and `textarea`. I did not understand the extent of the issue, sorry.

christianlupus (approved):
Seems good enough to me as a fix. LGTM

PR #941, opened 2022-03-29
Correct spelling

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>

### DIFF
--- a/lib/Service/RecipeService.php
+++ b/lib/Service/RecipeService.php
@@ -1100,7 +1100,7 @@ class RecipeService {
 	public function getRecipeImageFileByFolderId($id, $size = 'thumb'): File {
 		$recipe_folders = $this->root->getById($id);
 		if (count($recipe_folders) < 1) {
-			throw new Exception($this->il10n->t('Recipe with id %d not found.', [$id]));
+			throw new Exception($this->il10n->t('Recipe with ID %d not found.', [$id]));
 		}
 		$recipe_folder = $recipe_folders[0];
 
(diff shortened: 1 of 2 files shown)


MarcelRobitaille (approved):

christianlupus (approved):
Looks good to me.

lib/Service/RecipeService.php
```diff
@@ -1100,7 +1100,7 @@ public function parseRecipeFile($file) {
 	public function getRecipeImageFileByFolderId($id, $size = 'thumb'): File {
 		$recipe_folders = $this->root->getById($id);
 		if (count($recipe_folders) < 1) {
-			throw new Exception($this->il10n->t('Recipe with id %d not found.', [$id]));
+			throw new Exception($this->il10n->t('Recipe with ID %d not found.', [$id]));
```
christianlupus, line 1103: @Valdnet i have one question regarding this change: is it only syntactical to have a better English writing or is this one a technical thing (do we have a duplicate strong here somewhere)

christianlupus:
If you do not mind, i will append a note to the Changelog. Your are free to do this yourself, i will need some time to get to my machine.

Valdnet:
> If you do not mind, i will append a note to the Changelog.

No problem, you can add 😉.

lib/Service/RecipeService.php
Valdnet: If the ID is a direct identifier and points to an individual, the item or item should be in uppercase.

PR #956, opened 2022-04-12
Do not throw 500 HTTP error if no user is logged in

### LINKED ISSUE #951: Error in UserConfigHelper, when accessing cookbook URL, while not being logged in
**Description**
Error in UserConfigHelper, when trying to access the cookbook URL, while not being logged in. A null object instead of a string is given to the constructor in this case.

**Reproduction**
1. Make sure to NOT being logged in at the Nextcloud instance.
2. Open the cookbook URL e.g., http://nextcloud.example.org/index.php/apps/cookbook/#/
3. There's a Nextcloud server error shown
4. In the Nextcloud log file, an error with the following message appears: Argument 1 passed to OCA\\Cookbook\\Helper\\UserConfigHelper::__construct() must be of the type string, null given

**Expected behavior**
Forward to the Nextcloud login page.

**Actual behavior**
Nextcloud shows a server error.

**Browser**
Firefox. The error should be independent of the browser, as it is a backend issue.

**Versions**
Nextcloud server version: 23.0.3
Cookbook version: 0.9.11
Database system: PostgreSQL

christianlupus:
Just a small regression in the dependency injection code.

Closes #951


seyfeb (approved):
LGTM

PR #973, opened 2022-05-02
Handle mandatory components in recipe JSON smoothly

### LINKED ISSUE #923: Unhelpful error message `Unknown answer returned from server. See logs.` given if recipe name missing
**Description**
When saving a recipe, the message `Unknown answer returned from server. See logs.` is shown if the name field is empty. Only the name field seems to be absolutely required. Adding `test` here allows the form to be submitted even if everything else is empty.

**Reproduction**
1. Click '+ Create Recipe'
2. Immediately click save
1. `Unknown answer returned from server. See logs.` is shown in a popup window

**Expected behavior**
There is a helpful message like `Title field is required`. This could even be shown inline. The input could get a red border and the text shown directly underneath.

**Actual behavior**
`Unknown answer returned from server. See logs.` is shown in a popup window

**Browser**
Firefox 97.0.1 on Linux

**Versions**
Nextcloud server version: Master branch
Cookbook version: Master branch (reproduced in my "production" instance running 0.9.10)

christianlupus:
Closes #923

This introduces a new possible status in the case of storing a recipe (both new and existing). If the name was empty, the recipe could not be stored and thus a 406 is returned by the server.

It can easily be extended for other mandatory entries in the recipe description.


christianlupus:
This PR just focuses on the backend part. The mentioned frontend issues as suggested in #923 with red border and subtext etc must be implemented in the frontend. This is considered an additional feature and should be handled in a dedicated feature request.

lib/Controller/RecipeController.php
```diff
@@ -100,7 +101,16 @@ public function update($id) {
...
-		$file = $this->service->addRecipe($recipeData);
+		try {
+			$file = $this->service->addRecipe($recipeData);
+		} catch (NoRecipeNameGivenException $ex) {
+			$json = [
+				'msg' => $ex->getMessage(),
+				'file' => $ex->getFile(),
+				'line' => $ex->getLine(),
+			];
+			return new JSONResponse($json, Http::STATUS_NOT_ACCEPTABLE);
```
seyfeb, line 112: Is this really the correct status code to return here? Isn’t this more like a 400?

src/components/RecipeEdit.vue
```diff
@@ -596,6 +596,7 @@ export default {
                         // Non 2xx state returned
 
                         switch (e.response.status) {
+                            case 406:
```
seyfeb, line 599: Maybe use different code?

lib/Controller/RecipeController.php
christianlupus: Yes, you are right. 400 seems more suited. I do not like the 400 as it is items used as a catch all code. But we do not use it currently as far as I know. 

I will update soon.

christianlupus: I fixed this.

src/components/RecipeEdit.vue
christianlupus: Fixed.

lib/Controller/RecipeController.php
seyfeb: I understood your intention, but looking through the 4xx error codes on https://developer.mozilla.org/en-US/docs/Web/HTTP/Status only 422 seemed suited but is restricted to WebDAV. we could misuse it though to have a more specific code...

seyfeb (approved):

lib/Controller/RecipeController.php
christianlupus: Would you prefer the 400 or the 422? I'd maybe even go for the 422 as we are building the API in any case. So I would maybe even vote for 422. I will wait with merging for your thoughts.

We can of course say that for that specific endpoint (`/api/recipe/<id>/image`) the 400 is sufficient and we provide an error message as well...

seyfeb: I’d support either way, take your pick ;)

PR #982, opened 2022-05-10
Allow the client to specify the image type using Accept header

### LINKED ISSUE #954: Recipe Image Endpoint should not return two different data types
As discussed in #418:

> @Teifun2 if this needs more discussion, we might be better of in a new issue instead of this PR.
> 
> Apart from that, can't you check the `Content-Type` header before?

The problem for me is that i am using Network Cached Image widgets. These use they own implementation of the network controller. I can define the behaviour if the recource is not available. However the svg return results in a lot of exceptions as it is accepted as an image but can then not be parsed.

In my opinion it would be better to store the svg in the frontend and dispaly it in case images cannot be loaded, and just return 404 on the endpoint if no recipe image is defined.

This would make it so that the endpoint itself does not change data types.

(I would be willing to try and implement this, if it would be an accepted behaviour!)

christianlupus:
Closes #954

This is a quick fix for issue #954. This checks the accept header and returns a 406 HTTP error in case no JPG file was found on the server,

This is not yet capable to support other image files (like PNG) but this is the first step toward multi-type image files. The multi-type must be implemented separately as it is a more involved change/refactor involved.

### DIFF
--- a/lib/Controller/RecipeController.php
+++ b/lib/Controller/RecipeController.php
@@ -14,8 +14,10 @@ use OCA\Cookbook\Service\RecipeService;
 use OCP\IURLGenerator;
 use OCA\Cookbook\Service\DbCacheService;
 use OCA\Cookbook\Exception\RecipeExistsException;
+use OCA\Cookbook\Helper\AcceptHeaderParsingHelper;
 use OCA\Cookbook\Helper\RestParameterParser;
 use OCP\AppFramework\Http\JSONResponse;
+use OCP\IL10N;
 
 class RecipeController extends Controller {
 	/**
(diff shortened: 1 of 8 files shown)


lib/Controller/RecipeController.php
```diff
@@ -178,16 +201,28 @@ public function destroy($id) {
...
 
 			return new FileDisplayResponse($file, Http::STATUS_OK, ['Content-Type' => 'image/jpeg', 'Cache-Control' => 'public, max-age=604800']);
 		} catch (\Exception $e) {
-			$file = file_get_contents(dirname(__FILE__) . '/../../img/recipe.svg');
-
-			return new DataDisplayResponse($file, Http::STATUS_OK, ['Content-Type' => 'image/svg+xml']);
+			if (array_search('svg', $acceptedExtensions, true) === false) {
+				// We may not serve a SVG image. Tell the client about the missing image.
+				$json = [
+					'msg' => $this->l->t('No image with the matching mime type was fount on the server.'),
```
seyfeb, line 217:
```suggestion
					'msg' => $this->l->t('No image with the matching mime type was found on the server.'),
```

lib/Helper/AcceptHeaderParsingHelper.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace OCA\Cookbook\Helper;
+
+/**
+ * This class parses the Accepts header from a HTTP request and returns an array of accepted file extensions.
+ *
+ * The return values is a list of extensions that the client is willing to accept.
```
seyfeb, line 8:
```suggestion
 * This class parses the Accepts header of an HTTP request and returns an array of accepted file extensions.
 *
 * The return value is a list of extensions that the client is willing to accept.
```

seyfeb (approved):

PR #988, opened 2022-05-13
l10n: Change to uppercase

rakekniven:
See https://en.wikipedia.org/wiki/MIME

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>


christianlupus (approved):
LGTM

PR #1007, opened 2022-05-18
Add a helper class to manage the user folders

### LINKED ISSUE #1004: The user folder is checked and (re) generated for one request once per recipe. This should be cached.
I have this already prepared as part of a draft PR. I just need to extract and integrate with other classes.

christianlupus:
Closes #1004 

This is a cherry-pick (plus modifications) from another dev branch to bring the functionality of a dedicated user folder helper class to the master branch. The class has some caching included to prevent unneeded calls to the DB.

This makes roughly 20% of the query time of fetching all recipes currently:
grafik


christianlupus:
@seyfeb The fixupbot seems to be broken at the moment. If you would confirm/approve the PR, I will force-merge it.

Issue #896, opened 2022-02-20
UI glitch with narrow devices


christianlupus:
**Description**
At least when viewing a recipe the navigation is not handling the options correctly.

**Screenshots**
![screenshot at desktop](https://user-images.githubusercontent.com/8202922/154853365-dc01c804-d32c-4abd-93b5-452b08442548.png)
![screenshot at narrow device](https://user-images.githubusercontent.com/8202922/154853348-55f6fe93-e3b9-4595-a8a6-6f447e8d427d.png)

The reload, edit and print actions are obviously not visible on the narrow device.

**Browser**
Which browser are you using? Firefox

MarcelRobitaille:
I would like to try this one. Should the icon buttons stay in the panel? It looks like there is space for them once the text goes into an expanding menu.

seyfeb:
I'd say it wouldn't hurt to have them directly accessible and not hidden behind an additional burger menu if there is sufficient space

MarcelRobitaille:
How about the folder icons with no text next to them in the screenshot? Is that part of this issue?

seyfeb:
I think this is related. If I remember correctly, the actions (print, delete, edit) were added as breadcrumbs. My guess is, that they are hidden within the dropdown as a folder icon with no text.

MarcelRobitaille:
I am not able to reproduce this unless I make my viewport extremely small (243px). Before that, the recipe title is still visible, so there is no space for the buttons in the right. 

@christianlupus, May I ask what your resolution is when experiencing this issue?

![image](https://user-images.githubusercontent.com/8503756/157088569-d3b4379e-2241-4a42-98f9-482b058f8356.png)

Here is a video with different sizes

https://user-images.githubusercontent.com/8503756/157096550-adcdb911-5de2-4d70-985d-0900f83a66e4.mp4

There also seems to be a related problem around 5 seconds where one button goes out of the viewport to the right. Should I open a new issue for this or do you think that this has the same root cause?

MarcelRobitaille:
I think `Breadcrumbs` may be the wrong component for this. From what I have seen, this is for displaying the ancestry of a folder. Here is the example from the docs:
![image](https://user-images.githubusercontent.com/8503756/157275690-37a4ec85-3336-46d3-9069-134e7cc214d3.png)

[...]

PR #1091, opened 2022-07-20
UI glitch with narrow devices

MarcelRobitaille:
Fixes: #896

Fix this UI glitch by replacing the [`Breadcrumbs`](https://nextcloud-vue-components.netlify.app/#/Components/Breadcrumbs?id=breadcrumbs-1) element with an element better suited for this kind of button layout. A `ResponsiveActions` was developed based on the [`nextcloud/vue Actions`](https://nextcloud-vue-components.netlify.app/#/Components/Actions?id=actions-1) component. This `ResponsiveActions` shows the buttons with an icon and text if there is enough space, only the icons if there is only enough space for that, and only a 3-dot menu if very narrow devices like phones. This is to avoid  [mystery meat navigation](https://en.wikipedia.org/wiki/Mystery_meat_navigation).

https://user-images.githubusercontent.com/8503756/172550938-d1241efe-cef2-487c-a0bf-79b4e26c2144.mp4


MarcelRobitaille:
Also, not sure how copying files from nextcloud/nextcloud-vue works. It's GPLV3, so we probably need to state the changes and disclose the source. I'm not sure how that works though, I've never done that before.

[`src/assets/variables.scss`](https://github.com/nextcloud/nextcloud-vue/blob/master/src/assets/variables.scss) and [`src/components/validateSlot.js`](https://github.com/nextcloud/nextcloud-vue/blob/master/src/utils/ValidateSlot.js) are copied unchaged. `ResponsiveActions` is based on [`Actions`](https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/Actions/Actions.vue).

christianlupus:
I just compiled it and found a small further glitch:
grafik

To reproduce:
1. Open recipe for editing. Refresh page to reload any JS code.
2. Scale window such that "folding" begins but text (barely) present
3. Save recipe causing more buttons to appear
4. Rescaling seems to use the with of the original buttons (2 buttons during save) as a boundary for further folding.

MarcelRobitaille:
Thanks for the bug report. I will try to reproduce and fix it.

I understand. I know licensing is tricky. I think it could be made into a PR. My issue requesting this upstream was completely ignored, but maybe a PR would be different.

MarcelRobitaille:
Closing this because I think we're going with #1105

PR #1099, opened 2022-07-23
Rename the app from nextcloud-cookbook to cookbook

MarcelRobitaille:
The old name causes problems when using the webpack `file-loader`. The
URL from this would start `/apps/nextcloud-cookbook` although the proper
install location and the proper URL to load the file is
`/apps/cookbook`.
Besides this `file-loader` issue, this change makes everything more
consistent and more conformal to best practices.

I am assuming that URLs like the GitHub repository and matrix should remain the same.


christianlupus (approved):
LGTM

PR #1119, opened 2022-07-26
Fix the endpoint used for updating recipes

### LINKED ISSUE #1073: Wrong API endpoints used for updating recipes from Vue
Currently there is a big in the Vue e frontend. Amy update on a recipe is directed to the engine for new recipes. This does work because the implementation in the backend are very similar. However, this should be fixed to prevent nasty bugs.

MarcelRobitaille:
Fixes #1073

`this.recipe_id` is null, so use the recipe ID saved in the store.

I've tested it in the browsed. Where I was seeing `POST` before, I am now seeing `PUT`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   [#1105](https://github.com/nextcloud/cookbook/pull/1105) @MarcelRobitaille
 - Fix typos and styling issues
   [#1112](https://github.com/nextcloud/cookbook/pull/1112) @christianlupus
+- Fix API endpoint used for updating recipes
+  [#1119](https://github.com/nextcloud/cookbook/pull/1119) @MarcelRobitaille
 
 ### Maintenance
 - Add composer.json to version control to have unique installed dependency versions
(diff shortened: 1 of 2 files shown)


christianlupus:
Hey. I just tried it and now everything gets done via `PUT`. A new recipe is created via `PUT` on `/apps/cookbook/api/recipes/0`.

Maybe you can have a second look.

MarcelRobitaille:
There I go being impatient again. Sorry about that. I think I got it now. I saw that line `if (this.$route.params.id)` elsewhere in the code. I actually tested it in both cases this time too. I got `PUT` for update and `POST` for create. BTW, I think if `id` can be 0, we should do the test `if (this.$route.params.id ?? false)`. This is checking that it's not `null` and not `undefined`, but can be 0, empty string, etc.

christianlupus:
I think all databases use a positive integer as primary key. Thus, it should work. For MySQL i am pretty sure and PostgreSQL i think all well. Only SQLite might cause trouble.

We can however ignore that. The id is the id of the parent folder at the moment. This parent folder will never be the very first entry in the list of notes present in the cloud. The skeleton files are created earlier. Thus, we can suspect them to be largen than 0.

christianlupus (approved):
LGTM

Issue #907, opened 2022-03-05
Blur/unfocus the background upon popup showing


christianlupus:
When a popup is shown like when user should select a recipe by typing the `#`, this popup has too l little contrast to the rest of the page. It is hard for the eye (at least for mine) to see that I need to do something in a dialog and should not use the rest of the page instead.

I would suggest to either blur the background (aka main app canvas), make it semi-transparent, put a semi-transparent black/gray filter over it or put a more prominent frame around the popup to get more prominent.

ananyak19:
Hey, is this available? 
Can I work on this?

MarcelRobitaille:
I think this issue should be reconsidered, at least for the sharp popup. The point of #1098 was to make that popup less obtrusive, so it should definitely not have a blurred background. Maybe for other popups this still makes sense, but if we use the proper Nextcloud Vue components, I think it handles that for us.

As far as being available, nobody is working on this at the moment as far as I know.

christianlupus:
We have only a handful of popups present ATM. The selection of recipes during the linking phase as described by @MarcelRobitaille I consider no longer a real popup.

For the other popups the Vue components of `@nextcloud/vue` have that already handled by the library. So, I would close this issue now as it is no longer a valid request.

PR #1122, opened 2022-07-27
Add an overlay when the app navigation is open

### LINKED ISSUE #1117: Navigation button is hard to see
When looking at the current state of #1105, the navigation menu is reappearing again. See below:
grafik

When the navigation is expanded and the zoom factor is set correctly, the button might overlap with the edit (or other) button(s).

At least for me, this is almost invisible. I have seen this problem already in the past also with other apps on my mobile that the styling of the nav buttons might conflict with underlying content. Graphical example:

grafik

I am not sure if this is really an app issue or better suited in the Nextcloud style guides (as other apps are affected similarly).

MarcelRobitaille:
Fixes #1117 

What do you think of this solution? The hamburger menu is still kind of hard to see just because it's gray on gray. I made it so clicking the overlay also closes the navigation.

The alternative that I see is to add our own button to open the navigation in the header and our own button to close the navigation in the navigation itself. We would have to hide the current button. We would also have to use the `isMobile` mixin to see if the button should be present.


seyfeb:
I'm not too deep in this, but just an idea: Is it possible to simply override the buttons css? E.g., add a semi-transparent white background to the button in the non-hover/non-active state? Just for demonstration, sth along the lines of:

```
.app-navigation > button {
  background-color: rgb(255, 255, 0.6);
  border-top-left-radius: 0;
  border-bottom-left-radius: 0;
}
```

Screenshot 2022-07-27 at 11 05 57

MarcelRobitaille:
@seyfeb That could work. We'd need some more logic to show the default gray background on hover/focus when not in the overlay mode.

image

I am leaning towards hiding the stock button and using one in the sidebar itself. It seems cleaner to me, and I've already ranted about the stock button, so I'll spare you from that.

christianlupus:
I consider the black/gray semi-transparent background a good solution like already suggested in #907 in another context. Apert, I am no friend of building too much blocks on our own duplicating the work done by the Nextcloud core team.

MarcelRobitaille:
@christianlupus thanks for the feedback

> Apert, I am no friend of building too much blocks on our own duplicating the work done by the Nextcloud core team.

I am open to suggestions for how to avoid this, but I don't really see any option. #1117 was opened because of a shortcoming of the core UI. The options I see are work around that or live with it. From what I can tell, #1117 is in some sense intended behaviour for nextcloud-vue. I see the same thing in other apps (e.g. calendar).

christianlupus (approved):
LGTM

PR #1145, opened 2022-08-07
Update npm and node engines versions

nextcloud-command:
Automated update of the npm and node engines versions


christianlupus (approved):
We need to update. Lets do this.

PR #1258, opened 2022-10-18
Create an app configuration

### LINKED ISSUE #1067: Create an app configuration
Currently, our app is configured in a small overlay in the lower-left corner of the screen.

As more and more requests related to configurable enhancements are on the bug tracker, it might be a considerable approach to put the settings into a full-frame page. This can be done in different ways:
- the per-user configuration and register with the NC core
- a custom setting page (overlay or similar) that can be made visible

In this issue, the usage of mobile devices should be made as an additional request: Any changes to the settings should be easily achievable also on a mobile screen with touch control.

MarcelRobitaille:
Fixes #1067

Move the AppNavigationSettings from the sidebar to a AppSettingsDialog modal that covers the screen.

This is what it should look like:

image

This makes it easy for us to add more configuration options for things like notifications and alarms without cluttering the sidebar.

### DIFF
--- a/src/components/SettingsDialog.vue
+++ b/src/components/SettingsDialog.vue
@@ -6,6 +6,7 @@
         first-selected-section="keyboard shortcuts"
     >
         <NcAppSettingsSection
+            id="settings-recipe-folder"
             :title="t('cookbook', 'Recipe folder')"
             class="app-settings-section"
         >
@@ -46,6 +47,7 @@
             </fieldset>
         </NcAppSettingsSection>
         <NcAppSettingsSection
+            id="settings-recipe-display"
             :title="t('cookbook', 'Recipe display settings')"
             class="app-settings-section"
         >
--- a/src/components/SettingsDialog.vue
+++ b/src/components/SettingsDialog.vue
@@ -1,19 +1,24 @@
 <template>
-    <NcAppNavigationSettings :open="true">
-        <div id="app-settings">
+    <NcAppSettingsDialog
+        :open.sync="isOpen"
+        :title="t('cookbook', 'Cookbook settings')"
+        :show-navigation="true"
+        first-selected-section="keyboard shortcuts"
+    >
+        <NcAppSettingsSection
+            :title="t('cookbook', 'Recipe folder')"
+            class="app-settings-section"
+        >
             <fieldset>
                 <ul>
                     <li>
-                        <NcActionButton
-                            class="button"
-                            :icon="
-                                scanningLibrary
-                                    ? 'icon-loading-small'
-                                    : 'icon-history'
-                            "
-                            :title="t('cookbook', 'Rescan library')"
-                            @click="$emit('reindex')"
-                        />
+                        <NcButton @click="$emit('reindex')">
+                            <template #icon>
+                                <LoadingIcon v-if="scanningLibrary" />
+                                <ReloadIcon v-else />
+                            </template>
+                            {{ t("cookbook", "Rescan library") }}
+                        </NcButton>
                     </li>
                     <li>
                         <label class="settings-input">{{
@@ -37,6 +42,15 @@
                             placeholder="0"
                         />
                     </li>
+                </ul>
+            </fieldset>
+        </NcAppSettingsSection>
+        <NcAppSettingsSection
+            :title="t('cookbook', 'Recipe display settings')"
+            class="app-settings-section"
+        >
+            <fieldset>
+                <ul>
                     <li>
                         <input
                             id="recipe-print-image"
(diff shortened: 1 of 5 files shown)


MarcelRobitaille:
@christianlupus That's weird, running `npm run styleling` locally doesn't warn about anything for me. I ran prettier, eslint, and styleling before pushing. BTW, I'm getting a deprecation warning with eslint now.

christianlupus:
Are you on the latest styleling config? There was a PR recently #1254.

src/components/SettingsDialog.vue
```diff
@@ -1,19 +1,24 @@
...
-                            class="button"
-                            :icon="
-                                scanningLibrary
-                                    ? 'icon-loading-small'
-                                    : 'icon-history'
-                            "
-                            :title="t('cookbook', 'Rescan library')"
-                            @click="$emit('reindex')"
-                        />
+                        <NcButton @click="$emit('reindex')">
```
christianlupus, line 15:
```suggestion
                        <Button @click="$emit('reindex')">
```

src/components/SettingsDialog.vue
```diff
@@ -1,19 +1,24 @@
...
+                            <template #icon>
+                                <LoadingIcon v-if="scanningLibrary" />
+                                <ReloadIcon v-else />
+                            </template>
+                            {{ t("cookbook", "Rescan library") }}
+                        </NcButton>
```
christianlupus, line 21:
```suggestion
                        </Button>
```

src/components/SettingsDialog.vue
```diff
@@ -64,35 +78,41 @@
...
     components: {
-        ActionButton,
-        AppNavigationSettings,
-    },
-    props: {
-        scanningLibrary: {
-            type: Boolean,
-            default: false,
-        },
+        NcButton,
```
christianlupus, line 101:
```suggestion
        Button,
```

src/components/SettingsDialog.vue
```diff
@@ -64,35 +78,41 @@
...
 </template>
 
 <script>
-import ActionButton from "@nextcloud/vue/dist/Components/ActionButton"
-import AppNavigationSettings from "@nextcloud/vue/dist/Components/AppNavigationSettings"
+import { subscribe, unsubscribe } from "@nextcloud/event-bus"
+
+import AppSettingsDialog from "@nextcloud/vue/dist/Components/AppSettingsDialog"
+import AppSettingsSection from "@nextcloud/vue/dist/Components/AppSettingsSection"
+import NcButton from "@nextcloud/vue/dist/Components/Button"
```
christianlupus, line 90:
```suggestion
import Button from "@nextcloud/vue/dist/Components/Button"
```

src/components/SettingsDialog.vue
christianlupus, line 15: The button is causing a JS exception on my machine when clicking it. Can you confirm?

christianlupus (changes requested):
The PR has some problems yet. The component `NcButton` does not yet exist. It is part of the `@nextcloud/vue` v6 package. We are stuck with v5. Like this, the button is non-functional.

When changing the button, I get an error in the JS console:
grafik

Additionally, I cannot change the folder of the cookbook. I cannot navigate in the modal dialog to select a folder.

Can you confirm these issues?

MarcelRobitaille:
@christianlupus I called it `NcButton` and not `Button` because I got an error something like `Button is a reserved word in HTML`. There shouldn't be a problem with `import NcButton from '@nextcloud/vue/dist/Components/Button`, is there? It's fewer things to change when we finally migrate too.

I will have a look at resolving the other problems

MarcelRobitaille:
Here is the error when renaming `NcButton` to `Button`:

```
/home/marcel/code/cookbook/src/components/SettingsDialog.vue
  101:9  error  Name "Button" is reserved in HTML  vue/no-reserved-component-names
```

I think that is the cause of the errors in the screenshot you posted.

christianlupus:
> @christianlupus I called it `NcButton` and not `Button` because I got an error something like `Button is a reserved word in HTML`. There shouldn't be a problem with `import NcButton from '@nextcloud/vue/dist/Components/Button`, is there? It's fewer things to change when we finally migrate too.

Ahh, I am sorry, I missed that line. I assumed you were using `import {NcButton} from ...`. You are perfectly fine with that. Sorry for bothering you.

> Thanks for reviewing smile.
You are welcome.

Please tell me when you finished with the entire PR. Thanks!

MarcelRobitaille:
No worries.

I fixed the problem with the reindex button.

I am not sure what is going on with the file picker. I added this code in the developer console, and once the filepicker dialog is open, no events are fired at all:

```js
document.addEventListener('click', e => console.log(e))
```

I'm not really sure where this global `OC` is coming from. I can't find any documentation for it. I don't really like this kind of magic API, I prefer to import what I'm using explicitly. It's different from [this](https://github.com/nextcloud/nextcloud-dialogs/blob/master/lib/filepicker.ts), right?

I was not able to reproduce the error in your screenshot. Did you use the code exactly as it is here or did you replace `NcButton` with `Button` locally?

src/components/SettingsDialog.vue
MarcelRobitaille: I think this should be fixed in c9d37206. I had to replace `$emit('reindex')` with `reindex` because now that method is in the same component.

christianlupus:
> I'm not really sure where this global `OC` is coming from. I can't find any documentation for it. I don't really like this kind of magic API, I prefer to import what I'm using explicitly. It's different from [this](https://github.com/nextcloud/nextcloud-dialogs/blob/master/lib/filepicker.ts), right?

I do not know for sure. It comes from [here](https://github.com/nextcloud/cookbook/blob/296891ebf34f8b35fc691cd9cd0882f1b89a4afd/src/main.js#L38). I have read of it in the past but I do no longer find it.

> I was not able to reproduce the error in your screenshot. Did you use the code exactly as it is here or did you replace `NcButton` with `Button` locally?

Yes, I changed that locally. I have to retest it.

MarcelRobitaille:
I found the function: https://github.com/nextcloud/server/blob/master/core/src/OC/dialogs.js#L229-L572

It seems like a bug, but I'm not sure whether to report it to @nextcloud/vue or nextcloud/server.

MarcelRobitaille:
After hours of debugging, I found that this bug is not present in @nextcloud/vue <= 5.3.1. We could downgrade to this version until we move to 7.0.0, or we could try to upstream a patch and have it backported to 5.4.1.

MarcelRobitaille:
My guess is that the cause is https://github.com/nextcloud/nextcloud-vue/pull/2654. It seems like it was already [fixed in later versions](https://github.com/nextcloud/nextcloud-vue/pull/3020). Should we request a backport?

christianlupus:
I think a backport might be the cleanest solution.

Let's see, if @nextcloud/vuejs do know something about this issue. Can anyone of you confirm that there might be a problem and there is in fact a chance this can be fixed with the 5.x branch?

christianlupus:
ATM there are some commits that need fixing. I will wait and set this PR to be a draft. You can mark as ready for review once, you are done, @MarcelRobitaille.

MarcelRobitaille:
I rebased the NC25 changes. Now we're getting hung up on this open issue: https://github.com/nextcloud/nextcloud-vue/issues/3386

christianlupus:
:+1: for the rebase, :-1: for the upstream bug.

MarcelRobitaille:
The upstream issue was closed, but the fixes still haven't made it into a stable version.

MarcelRobitaille:
nextcloud/vue 7.10.0 was recently released, which fixes the "too much recursion" issue. I've just tested it and now everything seems to be working. The settings modal works and I can interact with the "Recipe folder" modal.

src/components/SettingsDialog.vue
christianlupus: Does no longer happen on my machine. So it seems to be fixed in the meantime.

christianlupus (approved):
Seems good to me now. LGTM

Issue #1257, opened 2022-10-17
Check every API request for correct content type


kolaente:
**Description**
Opening the cookbook app in a browser shows a loading spinner at "categories", does not seem to load anything (from the network tab and completely freezes the browser to the point where Firefox would offer to kill the tab. This looks like it is running into some kind of endless loop.

**Reproduction**
1. Open the cookbook app in the browser

**Expected behavior**
App working normally like it used to.

**Actual behavior**
Not working, locks up browser

**Screenshots**

This is what that looks like when it locks up:

![image](https://user-images.githubusercontent.com/13721712/196222408-86907458-d1a4-49c4-a6c2-302bc34450a2.png)

**Browser**

Tested in Firefox 104.0.2 and Chromium 105.0.5195.125, both on Linux (NixOS).

**Versions**
Nextcloud server version: 24.0.1
Cookbook version:  0.9.15
Database system: MariaDB

christianlupus:
Can you please open the developer tools before navigating to the cookbook app? There should be a few requests running. Potentially, the JS is not loading successfully. Could you please check this?

kolaente:
It looks like all assets are loaded successfully.

MarcelRobitaille:
Have you checked specifically for `cookbook-main.js`? Nextcloud itself loads many assets outside of this app, so it may not be obvious whether this app's assets are loaded. In the network tab, you can filter for `cookbook`, which for me shows that `cookbook-main.js` is being loaded (I have highlighted it in the screenshot).

![image](https://user-images.githubusercontent.com/8503756/196548158-18da8c56-9fa2-438c-a43c-f9af9262afc9.png)

MarcelRobitaille:
@kolaente Are there any errors in the console?

kolaente:
Only CSP warnings, not sure if they are related:

![image](https://user-images.githubusercontent.com/13721712/196639657-5ded5f6a-804d-43ca-80a8-96ef5fde1d0d.png)

MarcelRobitaille:
@kolaente Thanks. No, I also always get those. I don't think it's that.

>  Not working, locks up browser

What do you mean by this? Does the entire browser become unresponsive? Are other tabs and the browser's buttons also misbehaving / responding slowly?

kolaente:
> Does the entire browser become unresponsive? Are other tabs and the browser's buttons also misbehaving / responding slowly?

[...]

PR #1283, opened 2022-10-24
 Add logging to diagnose bugs in production

### LINKED ISSUE #1275: Add logging to diagnose bugs in production
- [x] Setup logging itself (maybe [winston](https://www.npmjs.com/package/winston))
- [x] Make log level configurable via browser console (this way users can enable logging in production, even if the app and settings UI isn't loaded)
- [x] Make the log level reset itself after some delay so users are not stuck with verbose logging forever
- [x] Add logs to the code

MarcelRobitaille:
Fixes #1275 

Add [a logging library](https://www.npmjs.com/package/vuejs-logger) and make enabling/disabling the logging and changing the log level configurable from localStorage. This way, we can instruct users to enable logging even before loading any of our JavaScript. This is very useful to troubleshoot issues that prevent our page from loading properly. The buttons to enable logging would obviously not work in this case.

To enable logging, a user only has to run the following before loading the app/refreshing the page:
```js
localStorage.setItem('COOKBOOK_LOGGING_ENABLED', 'true')
```

This will automatically be reset after 30 minutes so verbose logging is not enabled permanently for the user.

The log level is also configurable. For example:
```js
localStorage.setItem('COOKBOOK_LOGGING_LEVEL', 'debug')
```

I added some logs throughout the app, but I suspect we will flush this out as required. I also added some logs specific to troubleshooting #1257.

To add a log inside a Vue component, you only have to do:
```js
this.$log.debug('message')
```

To add a log outside a view component, you can do:
```
Vue.$log.debug('message')
```


christianlupus:
If possible fix the issues with prettier and add a short section to the dev documentation on how to enable the logger.

I have not tested the code though.

christianlupus (approved):
LGTM

PR #1291, opened 2022-10-26
Add debug logging on every network request

MarcelRobitaille:
This was requested in #1257

This basically adds an [interceptor](https://www.npmjs.com/package/axios#user-content-interceptors) to axios that logs every network request, which could be useful for debugging. It also checks the content-type of each request, and warns if it's not `application/json` or `text/json`.


christianlupus (approved):
LGTM

PR #1314, opened 2022-11-01
Create Release 0.10.0-rc1

christianlupus:
Create a rc for the NC25-compatible version 0.10.0


src/components/AppControls/AppControls.vue
```diff
@@ -322,7 +347,7 @@ export default {
     z-index: 2;
 
     /* The height of the nextcloud header */
-    top: var(--header-height);
+    top: 0px;
```
MarcelRobitaille, line 350: Maybe this can be completely removed at this point?

christianlupus: Then, the element is no longer sticky.

MarcelRobitaille: You're right. My bad

PR #1404, opened 2022-12-29
Make info block visibility configurable

### LINKED ISSUE #906: Make info blocks configurable, whether displayed or not displayed
It would be nice to be able to configure which info blocks should be displayed on the recipe and which should not. 
E. g. the info blocks Preparation time (hours:minutes); Cooking time (hours:minutes); Total time (hours:minutes); Nutrition Information; Tools are not interesting for me and distract from important (for me) part. 

The configuration should be global and not depending on the recipe

---

Depends on #1067

MarcelRobitaille:
Fixes #906 

- [x] Add the checkboxes in the settings UI
- [x] Store the settings
- [x] Use the settings to conditionally render the info blocks

### DIFF
--- a/src/components/SettingsDialog.vue
+++ b/src/components/SettingsDialog.vue
@@ -88,9 +88,14 @@
         >
             <fieldset>
                 <legend class="settings-info-blocks__legend">
-                    Control which blocks of information are shown in the recipe
-                    view. If you do not use some features and find them
-                    distracting, you may hide them.
+                    {{
+                        t(
+                            "cookbook",
+                            `Control which blocks of information are shown in
+                            the recipe view. If you do not use some features
+                            and find them distracting, you may hide them.`
+                        )
+                    }}
                 </legend>
                 <ul>
                     <li>
--- a/src/components/SettingsDialog.vue
+++ b/src/components/SettingsDialog.vue
@@ -91,9 +91,7 @@
                     {{
                         t(
                             "cookbook",
-                            `Control which blocks of information are shown in
-                            the recipe view. If you do not use some features
-                            and find them distracting, you may hide them.`
+                            "Control which blocks of information are shown in the recipe view. If you do not use some features and find them distracting, you may hide them."
                         )
                     }}
                 </legend>
--- a/src/components/SettingsDialog.vue
+++ b/src/components/SettingsDialog.vue
@@ -81,6 +81,69 @@
                 </ul>
             </fieldset>
         </NcAppSettingsSection>
+        <NcAppSettingsSection
+            id="settings-info-blocks"
+            :title="t('cookbook', 'Info blocks')"
+            class="app-settings-section"
+        >
+            <fieldset>
+                <legend class="settings-info-blocks__legend">
+                    Control which blocks of information are shown in the recipe view. If you do not use some features and find them distracting, you may hide them.
+                </legend>
+                <ul>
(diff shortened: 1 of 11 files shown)


MarcelRobitaille:
https://user-images.githubusercontent.com/8503756/210008931-daf84d45-3625-442d-8a09-0e911f5f1a7e.mp4

src/components/SettingsDialog.vue
```diff
@@ -81,6 +81,86 @@
...
+            <fieldset>
+                <legend class="settings-info-blocks__legend">
+                    {{
+                        t(
+                            "cookbook",
+                            `Control which blocks of information are shown in
+                            the recipe view. If you do not use some features
+                            and find them distracting, you may hide them.`
+                        )
+                    }}
```
christianlupus, line 98: This will not be translatable.

src/components/SettingsDialog.vue
```diff
@@ -218,20 +338,26 @@ export default {
...
+                    config.visibleInfoBlocks
+                )
                 this.showTagCloudInRecipeList =
                     this.$store.state.localSettings.showTagCloudInRecipeList
                 this.updateInterval = config.update_interval
                 this.recipeFolder = config.folder
-            } catch {
+            } catch (err) {
+                // eslint-disable-next-line no-console
+                console.error("Error setting up SettingsDialog", err)
```
christianlupus, line 360: Is this not better done using our logging mechanism?

tests/Unit/Controller/Implementation/ConfigImplementationTest.php
```diff
@@ -88,6 +88,7 @@ public function testList(): void {
 			'folder' => $folder,
 			'update_interval' => $interval,
 			'print_image' => $printImage,
+			'visibleInfoBlocks' => array(),
```
christianlupus, line 91: This is not sufficient to test it on PHP level. Please add a few test cases in the data provider at line 147.

christianlupus (changes requested):
Some minor changes are still required. Then, it can be merged.

src/components/SettingsDialog.vue
MarcelRobitaille: Because of the template literal? I was trying to avoid going over the line length.

src/components/SettingsDialog.vue
MarcelRobitaille: I guess I wanted to be consistent with the rest of the file. Shall I convert them all to use the logging or only this instance?

src/components/SettingsDialog.vue
christianlupus: The template literal is no problem as far as i know.

The thing is that the translation ci script users a regex that does not allow newlines. So, the complete string must be in one line. Otherwise, the string cannot be translated.

One has to add an exception for prettier to avoid the line to be changed. 

I asked for a fox but they fear to break other apps.

src/components/SettingsDialog.vue
christianlupus: As you like. I would at least fix lines that are changed anyways.

src/components/SettingsDialog.vue
MarcelRobitaille: Thank you for clarifying. I hope it is fixed.

christianlupus (approved):

PR #1405, opened 2022-12-29
Add network logging for responses, not only requests

### LINKED ISSUE #1185: Add client-side logging
Currently, we have to debug all problems on the client side manually, if we cannot reproduce in the dev environment. It would be nice to be able to enable verbose logging of all API requests in the browser logs including request data and answers. That should make it easier to get the needed informations.

---

Depends on #1067

MarcelRobitaille:
Fixes #1185 

Related to #1283 

Adds logging for network responses as well as network requests. With this, all the features requested in #1185 are implemented.

### DIFF
--- a/src/js/api-interface.js
+++ b/src/js/api-interface.js
@@ -10,17 +10,28 @@ const baseUrl = `${generateUrl("apps/cookbook")}/webapp`
 // Add a debug log for every request
 instance.interceptors.request.use((config) => {
     Vue.$log.debug(
-        `Making "${config.method}" request to "${config.url}"`,
+        `[axios] Making "${config.method}" request to "${config.url}"`,
         config
     )
     const contentType = config.headers[config.method]["Content-Type"]
     if (!["application/json", "text/json"].includes(contentType)) {
         Vue.$log.warn(
-            `Request to "${config.url}" is using Content-Type "${contentType}", not JSON`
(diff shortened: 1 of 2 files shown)


src/js/api-interface.js
```diff
@@ -10,18 +10,23 @@ const baseUrl = `${generateUrl("apps/cookbook")}/webapp`
...
         )
     }
     return config
 })
 
+instance.interceptors.response.use(
+    (response) => Vue.$log.debug("[axios] Received response", response),
+    (error) => Vue.$log.warn("[axios] Received error", error)
+)
+
```
christianlupus, line 29: This causes me a problem during loading the app. If I disable these lines, I can access the cookbook again.

The following at least solved the problem:
```suggestion
instance.interceptors.response.use(
    (response) => {
        Vue.$log.debug("[axios] Received response", response)
        return response
    },
    (error) => Vue.$log.warn("[axios] Received error", error) /* ... */
)

```

I fear this is a problem with promises and the modification of mine does not cover the error case.

christianlupus (changes requested):
This change causes problems when loading the app. The categories and tags are no longer loaded successfully.

src/js/api-interface.js
MarcelRobitaille: I'm sorry. I have a bad habit of not testing thoroughly enough.

christianlupus (approved):

PR #1445, opened 2023-01-15
Add operation ids to external API description

### LINKED ISSUE #1421: API code generation enhancement
It's very good having an openapi spec to work with although it's not quite ready for automated library generation.
A list of problems I found:

- Some Objects returned do not have names like the [Category get](https://github.com/nextcloud/cookbook/blob/4222b4815505ad913063f46c72647dd583751f58/docs/dev/api/0.1.0/openapi-cookbook.yaml#L415-L434) for example (see matrix discussion)
- missing `operationId` on the endpoints

Those will lead to the code generator deciding on the object name or the function name to invoke a call. And it gets ugly very fast like `ApiV1KeywordsGet200ResponseInner`or `ApiV1CategoryCategoryGetRequest`.

- incorrect use of types like [`string|integer`](https://github.com/nextcloud/cookbook/blob/4222b4815505ad913063f46c72647dd583751f58/docs/dev/api/0.1.0/objects.yaml#L349) (wich should be `oneOf: ...`) or plain [String for date-time](https://github.com/nextcloud/cookbook/blob/4222b4815505ad913063f46c72647dd583751f58/docs/dev/api/0.1.0/objects.yaml#L334-L335) (open PR in #1419)

The first is strictly an invalid openapi spec while the second one would lead to the generated lib handling the validation and parsing of the date format into a suitable object to use in the code.

All of the above would not need an altering of the api itself but a change in the documentation. The added properties would allow better and more efficient code generation and better openapi complience.

I'd like to add solutions to my already open PR and discuss everything over there although I'd need some help in finding good names for the `operationId` field and I'm quite bad at picking them :)

As mentioned the API itself wouldn't change but people already using the current spec for code generation would need to adapt to it. I don't think this would be a problem as the migration would be marginal (no logic change just replacing the object name `crtl+f` and replace would do the trick). Client authors could also just stay with the old spec as it would still be api compatible so no code change required at all.

null

christianlupus:
Closes #1421


christianlupus:
@Leptopoda, please have a look at this PR if that is as you think it useful.

Leptopoda:
I don't think underscrore is the right way to go.
I can only speak for dart/flutter but I think the general codestyle is the same on java/kotlin aswell.

Depending on the generator used it will generate something like:

```dart
final API _api = API();
final CategoriesApi _categoryAPI = _api.categories;

final Response<RecipeStub> _response= await _categoryAPI.recipes_in_category(id: "12345678");

if(response.code == 200){
  print(response.data)
}
```

In this case having `category` in the description is a bit verbose as we call a method to get the recipes on the categoryAPI object anyways. Also this does not easily resolve the http method used. 

Taking those and the aforementioned camelcase into consideration I'd name this particular endpoint `getRecipes`or `getRecipeStub` resulting in something like:

```dart
final _api = API();
final _categoryAPI = _api.categories;

final _response = await _categoryAPI.getRecipes(id: "12345678");

if(response.code == 200){
  print(response.data)
}
```

As dart can do type inference I could even leave out the type declarations as seen above.

---

Obviously I'm biased and prefer the style of the language I like/use most but I think the same should at least apply to java/kotlin. (iirc the preferd method naming scheme is camelcase and at least the official swagger kotlin generator does produce objects for each api endpoint).

---

Finally I've also found a few other problems with the spec that I'd like to discuss as they might also need server side changes while hopefully not being considered breaking. Therefore I'd not close the issue #1421 and change the PR content.

I've experimented with some changes already and played arround with them locally so theres nothing final or PR ready as of now (the reason I didn't go forward with them yet) but I'll attach my current version below.
I mainly changed some response objects to be named (the main reason this discuaaion started) and yes I've made the id an integer instead of a String but my generator does not like it the other way and as said this is still a WIP for me :)

[objects.txt](https://github.com/nextcloud/cookbook/files/10420171/objects.txt)
[openapi-cookbook.txt](https://github.com/nextcloud/cookbook/files/10420172/openapi-cookbook.txt)

christianlupus:
> I don't think underscrore is the right way to go. I can only speak for dart/flutter but I think the general codestyle is the same on java/kotlin aswell.

Yes, it is true that PascalCase and camelCase are the typical stylings for many modern languages. However, there are case-insensitive languages and these fail at either of these. Using kebab-case works well in YAML but might cause trouble with code generation (the minus might be a subtraction sign).
This is the reason, I went for snake_case.

In general, this can be discussed if we keep the names unique with case-insensitive comparisons.

> [...]
> In this case having `category` in the description is a bit verbose as we call a method to get the recipes on the categoryAPI object anyways. Also this does not easily resolve the http method used.

Here there is no real alternative. According to [the official documentation](https://swagger.io/docs/specification/paths-and-operations/#operationid) the operation ID must ba a _unique_ string.

As we have a way to get all recipes for a keyword, for a category, in a search and all recipes unconditionally, we must use appropriate names for these.
This means that the duplication of context is mandatory in some sense by the OpenAPI standard.

> Finally I've also found a few other problems with the spec that I'd like to discuss as they might also need server side changes while hopefully not being considered breaking. Therefore I'd not close the issue #1421 and change the PR content.
> 
> I've experimented with some changes already and played arround with them locally so theres nothing final or PR ready as of now (the reason I didn't go forward with them yet) but I'll attach my current version below. I mainly changed some response objects to be named (the main reason this discuaaion started) and yes I've made the id an integer instead of a String but my generator does not like it the other way and as said this is still a WIP for me :)

Let's focus here on the operation ID in this PR.

Leptopoda:
> Here there is no real alternative. According to the official documentation the operation ID must ba a unique string.

Yeah I forgot about that :(
also discussed here: https://github.com/OAI/OpenAPI-Specification/issues/381
We'll need to also include that information.

> Yes, it is true that PascalCase and camelCase are the typical stylings for many modern languages.

Can you please name me any relevant one? Also I can't imagine someone writing an API client in languages like Pascal :)
And if my 5 second search isn't wrong all [for us] relevant languages (Obj-C/Swift, php, JS/TS, Java/kotlin, dart) are both case sensitive and do also prefer camelCase for their functions.

Also a look at the languages supported by the official swagger/openapi codegenerator reveals that case insensitivity probably isn't a Problem.
> ActionScript, Ada, Apex, Bash, C, C#, C++, Clojure, Crystal, Dart, Elixir, Elm, Eiffel, Erlang, Go, Groovy, Haskell, Java, Jetbrains HTTP Client, Julia, k6, Kotlin, Lua, Nim, Node.js/JavaScript, Objective-C, OCaml, Perl, PHP, PowerShell, Python, R, Ruby, Rust, Scala, Swift, Typescript 

I could only imagine things like Haskell being problematic as they require method names starting with a capital letter (AFAIK).
Finally again: who is expected to write a client in Haskell? Also the generator should usually be smart enough to handle this stuff.

PR #1459, opened 2023-01-22
Use new logging framework consistently

MarcelRobitaille:
Replace all instances of `console.log` and alternatives with `this.$log.info` or `Vue.$log.info`.

In #1404, I noticed that although a logging framework was added in #1283, `console.log` and `console.error` were still used frequently throughout the app. This PR aims to replace all of these with the logging framework.

### DIFF
--- a/src/components/AppMain.vue
+++ b/src/components/AppMain.vue
@@ -48,7 +48,7 @@ export default {
     watch: {
         /* This is left here as an example in case the routes need to be debugged again
         '$route' (to, from) {
-            console.log(this.$window.isSameBaseRoute(from.fullPath, to.fullPath))
+            this.$log.debug(this.$window.isSameBaseRoute(from.fullPath, to.fullPath))
         },
         */
     },
--- a/src/components/AppNavi.vue
+++ b/src/components/AppNavi.vue
@@ -324,8 +324,7 @@ export default {
                             continue
                         }
                         if ($this.$refs[`app-navi-cat-${i}`][0].opened) {
-                            // eslint-disable-next-line no-console
-                            console.log(
+                            this.$log.info(
                                 `Reloading recipes in ${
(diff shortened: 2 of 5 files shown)


MarcelRobitaille:
Some `console.log`s are inside `if (this.verboseDebugLogging)`. Can this logic be removed now that #1283 has functionality to enable and disable logging?

src/components/AppMain.vue
```diff
@@ -48,7 +48,7 @@ export default {
     watch: {
         /* This is left here as an example in case the routes need to be debugged again
         '$route' (to, from) {
-            console.log(this.$window.isSameBaseRoute(from.fullPath, to.fullPath))
+            this.$log.debug(this.$window.isSameBaseRoute(from.fullPath, to.fullPath))
         },
         */
```
christianlupus, line 53: These lines are commented out. After it is redirected to a debug-level logger, we could remove the comment and just let the logger ignore the logged data in the normal (non-debug) case. Am I right?

christianlupus:
> Some `console.log`s are inside `if (this.verboseDebugLogging)`. Can this logic be removed now that #1283 has functionality to enable and disable logging?

Yes, I would say so, you can switch to the new logger with debugging level. This seems a good idea.

A general question: What do you consider needed to finish this PR?

src/components/AppMain.vue
MarcelRobitaille: Sure. Sounds like a plan.

MarcelRobitaille:
I was just waiting for your feedback on how to handle `verboseDebugLogging`

christianlupus:
Ahh, ok. Sorry for the delay.

MarcelRobitaille:
No worries. I've been super busy too.

christianlupus (approved):

PR #1482, opened 2023-02-09
Feature/display ratings

jmta:
Addition of displaying the aggregate rating of a recipe where available.

Ensured nothing is presented when no rating is available

If this is too small a change to be worth a  pull request. I'll wait until I've got additional elements working in a neat way and raise another PR when appropriate.

### DIFF
--- a/src/components/RecipeView.vue
+++ b/src/components/RecipeView.vue
@@ -67,6 +67,13 @@
                         <strong>{{ t("cookbook", "Servings") }}: </strong
                         >{{ $store.state.recipe.recipeYield }}
                     </p>
+                    <p v-if="$store.state.recipe.aggregateRating.ratingValue != null">
+                        <strong>{{ t("cookbook", "Rating") }}: </strong
+                        >
+
+                        <span v-for="i in parseInt($store.state.recipe.aggregateRating.ratingValue)" class="icon-favorite rating-icon" />
+                        <span v-for="i in 5-parseInt($store.state.recipe.aggregateRating.ratingValue)" class="icon-favorite-white rating-icon" />
+                    </p>
                 </div>
                 <div class="times">
                     <RecipeTimer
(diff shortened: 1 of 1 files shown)


src/components/RecipeView.vue
```diff
@@ -67,6 +67,12 @@
                         <strong>{{ t("cookbook", "Servings") }}: </strong
                         >{{ $store.state.recipe.recipeYield }}
                     </p>
+                    <p v-if="$store.state.recipe.aggregateRating != null">
+                        <strong>{{ t("cookbook", "Rating") }}: </strong
+                        >
+                        <span v-for="i in parseInt($store.state.recipe.aggregateRating.ratingValue)" class="icon-favorite rating-icon" />
+                        <span v-for="i in 5-parseInt($store.state.recipe.aggregateRating.ratingValue)" class="icon-favorite-white rating-icon" />
```
seyfeb, line 74: You might want to use mdi components for the icons as demonstrated [here](https://nextcloud-vue-components.netlify.app/#/Components/NcActions).
Also you could add half-star icons for values near  `n.5`. E.g., `n.25` < `rating` < `n.75` has  `n + a half` star?

seyfeb (commented):
Does it make sense to show the aggregate rating when there is no (user-friendly and sensible) way to manipulate it?

src/components/RecipeView.vue
jmta: Ah hadn't seen that before, I'll dive into that now. I'm going to have to resubmit this PR anyway as I've missed the sign off, and I now have 5 commits for as many lines of code I've added which is just unneeded.

jmta:
> Does it make sense to show the aggregate rating when there is no (user-friendly and sensible) way to manipulate it?

I'm unsure on the development ideologies on nextcloud/cookbook. While it might not be the full features many had hoped for its a step on the right path. If this is not what is felt necesarry I'm happy to save a few more RPs up.

PR #1541, opened 2023-03-17
Fix bugs in openapi 0.1.1

christianlupus:
This fixes some issues in the OpenAPI specifications after rework of it.


christianlupus:
@Leptopoda you think this is now better, right?

PR #1544, opened 2023-03-18
Removed obsolete parameters from example in OpenAPI spec

Leptopoda:
Probably an oversight of 07954bf2faffa5cd8bb54898a807680c06b532db

### DIFF
--- a/docs/dev/api/0.1.0/objects.yaml
+++ b/docs/dev/api/0.1.0/objects.yaml
@@ -235,7 +235,6 @@ Recipe:
       additionalProperties: true
   example:
     name: Baked bananas
-    "@type": "Recipe"
     id: "123"
     description: A very delightful recipe of the best baked bananas ever
     image: http://example.com/path/to/image.jpg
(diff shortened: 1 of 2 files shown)


Leptopoda:
The changelog already covers the api cleanups :)

christianlupus:
@Leptopoda, let me ask you one thing: Why would we want to remove the `@type` parameter from the API responses? I just checked that the API in fact returns the `@type` parameter. That way, we are compatible with the schema.org standard.

I am sorry, I did not get that merged earlier, and now I might ask dump questions that were mostly settled and I just forgot about it.

Leptopoda:
You removed it and I don't know why.

I only made the example reflect the specified responses (i.e. some documentation cleanup)

Leptopoda:
Also I should be fairly responsive so If you need a rebase I can always do so.

It might be better than you doing it so the commit is still properly signed.

christianlupus:
In that case, I will revert 07954bf and close this PR. It seems counter-intuitive to have implicit parts of the JSON objects. Sorry for the noise and doubled work. I am not sure why this was  done that way but it seems simply wrong.

PR #1550, opened 2023-03-20
Feature request: Main 0.9.x divide or multiply the list of ingredients

oxedgar:


christianlupus:
Let's keep the two development branches apart. 😁

PR #1582, opened 2023-03-27
Typo fix, also to prevent escaping issues

roliverio:
Replacing "Couldn't" with "Could not"

PR #1591, opened 2023-03-30
String fix

silopolis:
Non escaped single quote brakes visibleInfoBlocks() error alert modal text string


silopolis:
Closed, already fixed by #1582

PR #1700, opened 2023-05-28
Revert "Removed obsolete parameters from OpenAPI specs"

christianlupus:
## Topic and Scope

This reverts commit 07954bf2faffa5cd8bb54898a807680c06b532db.

The `@type` annotation should be still part of the OpenAPI description. Although the returned JSON objects are considered compatible with the schema.org standard, this does not mean, we should not make the implications explicit. Also, this makes reading more clean and avoid misunderstandings regarding if the `@type` is to be present or not.
## Concerns/issues

Just documentation, no effects

## Formal requirements


christianlupus:
This is a follow-up of #1544.

PR #1723, opened 2023-06-11
Enhance recipe recalculation algorithm

### LINKED ISSUE #1603: Enhance recipe recalculation algorithm
> I just checked a few additional cases. Now, I found a few issues, that might come up in the close future (or to help enhance the code base). I will nevertheless approve/merge this PR to get the functionality into the app. The fixes can then be applied step by step.

> I will open a new bug with all found issues in order to tackle them.

_Originally posted by @christianlupus in https://github.com/nextcloud/cookbook/issues/1573#issuecomment-1501198468_

The issues in question are the following:

- [x] The code assumes at least one whitespace char between the unit and the amount. E.g. in German, it is used like this `100g` without whitespace.
- [ ] Similarly, the code uses pure decimal points for fractions. In other languages, a comma is used as the decimal separator.
- [x] For large quantities, some users might add separators each 3 digits like this: `1,000`, `1 000`, or `1.000` This is especially critical as it conflicts with decimal comma above. I suggest to reject any number with both comma and point
- [ ] Fractions like `3 1/2 cups of sugar` will be parsed with a numerically wrong result. It should at least be rejected by the algorithm. However, it will even cause a wrong calculation without notification (ex: 2 x `1 1/2` should be 3, and the algorithm returns `2 1/2`). Better would be to parse these correctly.
- [x] I discovered rounding issues (6 servings used 2 eggs, changing to 9 servings displays `2.99` instead of the expected 3)
- [x] I would have preferred if the algorithm was encapsulated in its own module under `js` and imported from there. This might keep the overall code structure more tight and clear.

@j0hannesr0th what are you thinking about these points?

j0hannesr0th:
## Topic and Scope

Fixes at least partially #1603

See it live:

[Screencast from 11.06.2023 23:44:37.webm](https://github.com/nextcloud/cookbook/assets/36242595/c99b2463-cde2-406c-80b4-3b98ea8221aa)

## Concerns/issues

None

## Formal requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
   [#1573](https://github.com/nextcloud/cookbook/pull/1573) @j0hannesr0th
 - Add copy to clipboard action for ingredients
   [#1602](https://github.com/nextcloud/cookbook/pull/1602) @j0hannesr0th
+- Enhance recipe recalculation algorithm
+  [#1723](https://github.com/nextcloud/cookbook/pull/1723) @j0hannesr0th
 
 ### Fixed
 - Fix translation string to not contain quotes
--- a/src/components/RecipeView.vue
+++ b/src/components/RecipeView.vue
@@ -138,16 +138,17 @@
                         </template>
                         {{ t("cookbook", "Copy ingredients") }}
                     </NcButton>
-                    <h3 v-if="parsedIngredients.length">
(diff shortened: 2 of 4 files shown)


christianlupus (approved):
There are some flaws still in the code but this is yet much better than before.

I unlinked the underlying issue as it is not yet fully satisfied.

PR #1743, opened 2023-06-30
Fix missing parts of #1603

j0hannesr0th:
## Topic and Scope

Fix the problems described here: https://github.com/nextcloud/cookbook/issues/1603#issuecomment-1615002131

See it here:

[Screencast from 30.06.2023 23:38:52.webm](https://github.com/nextcloud/cookbook/assets/36242595/5c296507-a025-46b5-8b3d-b04d760ad0a8)

## Concerns/issues

@christianlupus what do you think about fraction handling like this?

I did some research and found this: https://github.com/infusion/Fraction.js/

I'm not sure if you want some 3rd party lib just for fraction handling.

## Formal requirements

- [X] I did check that the app can still be opened and does not throw any browser logs


christianlupus:
This was superseded by #1765

PR #1765, opened 2023-07-21
Fix/yield input calculator

christianlupus:
This supersedes #1603 which was closed by a bad git commit.


christianlupus:
Closes #1603

PR #1803, opened 2023-08-29
Fix/hooks

christianlupus:
This should fix some hooks and avoid recurring stash save/pop cycles that invalidate caches of the various tools involved. Instead a thin copy of the workspace is used in a temporary folder to run the checks.


christianlupus:
-approve

Issue #1641, opened 2023-04-30
tasty.co: "Could not parse recipe tools. It is no array."


Razzaline:
**Failing website:**
https://tasty.co/recipe/vegan-chocolate-chip-cookies

**Checking if valid metadata are present:**
Yes, I check the source code of the website and found metadata.

**Cookbook version:**
0.10.2

**Problem description (if applicable):**
The page cannot be imported. The error message is: "Could not parse recipe tools. It is no array."

PR #1864, opened 2023-10-29
WIP: Fix/single tool import support

seyfeb:
## Topic and Scope

Allows importing JSON-LD where the tools' field is a single tool instead of an array, like `tool: "My tool"` (see [#1641](https://github.com/nextcloud/cookbook/issues/1641))

## Formal requirements


CHANGELOG.md
```diff
@@ -27,6 +27,8 @@
   [#1834](https://github.com/nextcloud/cookbook/pull/1834) @christianlupus
 - Hode the button to copy ingredients unless there are some ingredients to copy
   [#1844](https://github.com/nextcloud/cookbook/pull/1844) @christianlupus
+- Allow single tool in JSON+LD import, fixes #1641
+  [#XXXX](https://github.com/nextcloud/cookbook/pull/1844) @seyfeb
```
christianlupus, line 31: This needs the correct pr number

christianlupus (approved):
Looks good in general. The issues in the code (see ci errors) need to be addressed, though 😁

PR #1892, opened 2023-11-15
Load config explicitly and remove setup routine of settings component

### LINKED ISSUE #1752: Configuration is stored upon each page visit
Currently, the config is loaded from the server upon loading *any* page (this is intended). Unfortunately, the code as provided in [the setup routine](https://github.com/nextcloud/cookbook/blob/3445f542a510b57e3e47dba2fc41837f92896a72/src/components/SettingsDialog.vue#L337-L362) detects a change in the internal structures and *stores* the same configuration to the backend again. That means that there are redundant calls and in the worst case, the configuration is messed up (with concurrent requests).

I think that there are multiple ways to enhance this structure:

1. The data should be provided by an initial state (provided by the server core) for the sake of reduced work. This avoids the additional request here to load any configuration.
2. The `setup` here is the wrong routine as it is intended to construct the Vue object. Instead, the data handling (fetching the config initially) should be done outside the component and the component should use the `mounted` method instead.
3. Accessing the global state might be better done with computed properties.
4. The update of configuration entries on the server might be more intelligent to only send a request if really needed.

@seyfeb the code basically comes from you. Am I misunderstanding anything here (especially point 2)?

christianlupus:
## Topic and Scope

This is a first step to prevent regular updating of the config on each page reload.

Closes #1752 

## Concerns/issues

The PR just covers the basic changes. The migration to the InitialState is not yet carried out.

## Formal requirements

### DIFF
--- a/src/components/SettingsDialog.vue
+++ b/src/components/SettingsDialog.vue
@@ -223,16 +223,10 @@ export default {
             isOpen: false,
             printImage: false,
             recipeFolder: "",
-            resetPrintImage: false,
             showTagCloudInRecipeList: true,
-            resetTagCloud: false,
             scanningLibrary: false,
-            // By setting the reset value initially to true, it will skip one watch event
-            // (the one when config is loaded at page load)
-            resetInterval: false,
             updateInterval: 0,
             visibleInfoBlocks: [...INFO_BLOCK_KEYS],
-            resetVisibleInfoBlocks: true,
             writeChanges: true,
         }
     },
@@ -242,11 +236,6 @@ export default {
                 return
             }
 
-            // Avoid infinite loop on page load and when reseting value after failed submit
-            if (this.resetPrintImage) {
-                this.resetPrintImage = false
-                return
-            }
             try {
                 await api.config.printImage.update(newVal)
                 // Should this check the response of the query? To catch some errors that redirect the page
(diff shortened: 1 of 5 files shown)


src/components/SettingsDialog.vue
```diff
@@ -276,40 +273,55 @@ export default {
...
-            if (this.resetVisibleInfoBlocks) {
-                this.resetVisibleInfoBlocks = false
+            if (!this.writeChanges) {
                 return
             }
+
             try {
                 const data = visibleInfoBlocksEncode(newVal)
                 await api.config.visibleInfoBlocks.update(data)
-                await this.$store.dispatch("refreshConfig")
```
seyfeb, line 292: This would immediately update the config in the store and should be maintained. Otherwise changing, e.g., the effect of the "Show cooking time" checkbox would only be visible once a recipe is reloaded and not when is is already visible.

src/components/SettingsDialog.vue
```diff
@@ -221,25 +223,19 @@ export default {
...
-            resetVisibleInfoBlocks: true,
+            writeChanges: true,
         }
     },
     watch: {
         async printImage(newVal, oldVal) {
-            // Avoid infinite loop on page load and when reseting value after failed submit
-            if (this.resetPrintImage) {
-                this.resetPrintImage = false
+            if (!this.writeChanges) {
```
seyfeb, line 235: We should dispatch the updated option to the store either here, or when settings dialog is closed (would be okay for the print option since its effect is not immediately visible on the screen.

seyfeb (approved):
Except for the two minor points it looks fine to me.

PR #1893, opened 2023-11-15
Switch to non-deprecated props

christianlupus:
## Topic and Scope

The `title` prop of the `NcAppNavigationItem` is deprecated and should be replaced by the `name` prop. This PR will do so.

## Concerns/issues

None known.

## Formal requirements


seyfeb (approved):
LGTM

Issue #752, opened 2021-07-01
Upgrade to Vue3


christianlupus:
Currently, we are running on Vue2. Once Vue3 is stable enough, we might want to switch to v3. This issue is merely a reminder.

Related are:
- #750 

Depends on #1079

christianlupus:
See especially the comments [1](https://github.com/nextcloud/cookbook/pull/916#issuecomment-1120229307) and [2](https://github.com/nextcloud/cookbook/pull/916#issuecomment-1120230164).

christianlupus:
This has been already solved and released

Issue #1891, opened 2023-11-15
Nasty flashing when finishing XHR requests


christianlupus:
The current implementation has multiple regions in the frontend that do either lazy loading or at least async loading. During the time of loading, some (more or less) static content is pre-rendered. This causes some flickering in the app when navigating around. I find it annoying and distracting, especially with slower network connections.

Where I have seen it yet (not complete list):

- [ ] Loading the home page first shows the no recipes found page then the list of recipes
- [ ] When viewing one recipe and then opening another one, the first is visible for a short amount of time
- [ ] Similar when editing a recipe

This is just visual but nevertheless annoying.

PR #1896, opened 2023-11-18
WIP: Prepare Vue.js 2 -> 3 upgrade

seyfeb:
## Topic and Scope

Soon an update to vue 3 is required. (#752)

This PR does some things in preparation:

- Bump dependencies
- Migrate all components to composition API
- Switch to `mitt` event bus library
- Restructure `.vue` files
- Change JS style to single quotes and required semicolons

## Formal requirements


christianlupus:
This PR is not yet working smoothly. I managed to get it on par with the current master branch at least such that it could be merged technically. Quite some of the changes made since the branch creation have been manually checked/incorporated by me. However, not everything is running smoothly.

I have seen two still present issues, maybe more are yet hidden as I just quickly tested the merged version:

1. The settings dialog should reaload the config.
2. The changed settings are for some reason not propagated to the store. This is only problematic for configs affecting the frontend but nevertheless...
3. While surfing, I had multiple errors in the browser console about wrong data types of various props. I have not looked into the details and I think, this are just small bugs but nonetheless...
grafik

5. #1891 is not fully addressed, just drop it from the list of solved issues, then we can fix it later.

For now, I am done (still feeling ill) and will have to rest a bit. Sorry, no better results.

christianlupus (approved):
Having clicked back and forth in the version, with the fixes applied, I see no breaks anymore in my dev environment.

PR #1942, opened 2023-12-06
fix: remove constant se in RecipeView

j0hannesr0th:
## Topic and Scope

@christianlupus there's a hardcoded "se" visible on all recipe pages

## Concerns/issues

none

## Formal requirements


christianlupus (approved):

christianlupus:
Uups, sorry for that.

PR #1944, opened 2023-12-06
fix: wrong parsing of recipe yield input

### LINKED ISSUE #1937: Changing number of servings not working after manual entry
**Description**
In general the new servings feature is working. But after entering the amount manually the increase/decrease buttons do not work anymore.

**Reproduction**
1. Load a receipt
2. Set the amount of servings to a number  (e.g. '5') by typing it manually
3. Click on the increase button (on the right)
4. Problem: The number '1' is concatenated to the previously added number (e.g. when entering '5' in step 2 the result wouldbe '51') 
5. Click in the decrease button (on the left)
6. Problem: The amount of servings field is empty

**Expected behavior**
When clicking on increase ('>') / decrease ('<') the number of servings should increase / decrease

**Actual behavior**
For '>' the number '1' is concatenated to the existing number
For '<' the amount of servings field gets empty

**Screenshots**
[nextcloud-cookbook-servings-bug.webm](https://github.com/nextcloud/cookbook/assets/1656210/5f33a32c-922e-477a-b8a3-ddafc4c8f753)

**Browser**
Which browser are you using?
Firefox 120.0b9
Chromium 119

**Versions**
Nextcloud server version: 25
Cookbook version: 0.10.4
Database system: MariaDB
OS: Linux Manjaro

j0hannesr0th:
## Topic and Scope

Fixes #1937

https://github.com/nextcloud/cookbook/assets/36242595/fa912d42-644f-40e9-b234-b32d3b83e2b7

## Concerns/issues

none.

## Formal requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
   [#1935](https://github.com/nextcloud/cookbook/pull/1935) @christianlupus
 - remove constant se in RecipeView
   [#1942](https://github.com/nextcloud/cookbook/pull/1942) @j0hannesr0th
-
+- fix wrong parsing of recipe yield input
+  [#1944](https://github.com/nextcloud/cookbook/pull/1944) @j0hannesr0th
 
 ## 0.10.3 - 2023-12-04
 
--- a/src/components/RecipeView/RecipeView.vue
+++ b/src/components/RecipeView/RecipeView.vue
@@ -80,10 +80,10 @@
                                         <span class="icon-view-previous" />
                                     </button>
                                     <input
-                                        v-model="recipeYield"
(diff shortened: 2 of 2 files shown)


christianlupus (commented):
Is it not simpler, to solve the root of the problem? In [line 83](https://github.com/nextcloud/cookbook/blob/8552ecbf102fd00f19f16495238341fe7cfe9489/src/components/RecipeView/RecipeView.vue#L83) the model is synced with the content of an input field. This causes the trouble of different data types. Normally, it is integer, but it is set to string internally after changes.

If you just added `.number.trim` (to `v-model`), the value is converted to a number before storing.

christianlupus (approved):

PR #1950, opened 2023-12-07
Formally make app compatible with NC28

christianlupus:
## Topic and Scope

Closes #1928

This marks the app as compatible with NC 28.

## Concerns/issues

No tests have been carried out yet.

## Formal requirements

There are some formal requirements that should be satisfied. Please mark those by checking the corresponding box.

- [ ] I did check that the app can still be opened and does not throw any browser logs
- [x] I created tests for newly added PHP code (check this if no PHP changes were made)
- [x] I updated the OpenAPI specs and added an entry to the API changelog (check if API was not modified)
- [x] I notified the matrix channel if I introduced an API change

PR #1963, opened 2023-12-12
Fix/1962 error when changing rescan period

### LINKED ISSUE #1962: Error when changing the rescan period
**Description**
In the settings an error message is shown when changing the rescan period.

**Reproduction**
1. Go to Settings
2. Delete number of rescan period.

**Expected behavior**
Since I am about to enter a different number, no error message should be shown.

**Browser**
Which browser are you using?

**Versions**
Nextcloud server version: 27, 28
Cookbook version: master, stable-0.10.x

seyfeb:
## Topic and Scope

Fixes the error message shown when the update interval field in the settings window is empty.

## Concerns/issues

Unfocusing the field does not show a warning, however then the value simply remains unchanged. In the future general form validation capabilities might be helpful (for the settings but also recipe editing), but afaik not supported by Nextcloud vue component by the time of writing.

Closes #1962 

## Formal requirements

Depends on #1950

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
   [#1944](https://github.com/nextcloud/cookbook/pull/1944) @j0hannesr0th
 - Update compatibility of app to NC28
   [#1950](https://github.com/nextcloud/cookbook/pull/1950) @christianlupus
+- **Settings:** Don't show error when update interval field is empty while typing.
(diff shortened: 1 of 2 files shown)


christianlupus:
Attention, this is based on the other PR #1950. So, do not merge yet but first handle the other PR.

christianlupus (approved):

PR #1977, opened 2023-12-13
Prevent searching in collapsed nutritient option field

### LINKED ISSUE #1969: Weird cursor in Select component after selecting option
> Weird cursor in Select component after selecting a value

I cannot check in NC27, as it was not working there as expected (no nutrition shown in FF on my dev instance right now)

Maybe you want to make the selection field unsearchable (`searchable="false"`)?

_Originally posted by @christianlupus in https://github.com/nextcloud/cookbook/issues/1950#issuecomment-1852040919_
            
This should be discussed and a solution be found.

Depends on #1950

christianlupus:
## Topic and Scope

Closes #1969

## Formal requirements


seyfeb (approved):
LGTM

PR #1990, opened 2023-12-14
Prevent ingredient warning for headings

### LINKED ISSUE #1985: Prevent ingredient warning for headings
> I find the quantity conversion of the ingredient fields very good. But: Can a code for the beginning of the line (e.g. the existing ##) be taken into account that this line is excluded from the recalculation, that this is a heading line or description line between the ingredients, for which there is also no error message that it cannot be converted?
Thanks!

_Originally posted by @neidert in https://github.com/nextcloud/cookbook/discussions/1561#discussioncomment-7765940_

j0hannesr0th:
## Topic and Scope

fixes #1985

https://github.com/nextcloud/cookbook/assets/36242595/4dd667b7-b7da-41c3-9de5-39f542c0dddd

## Concerns/issues

none.

## Formal requirements

### DIFF
--- a/src/components/RecipeView/RecipeIngredient.vue
+++ b/src/components/RecipeView/RecipeIngredient.vue
@@ -77,7 +77,6 @@ li {
 
 .header {
     position: relative;
-    left: -1.25em;
     margin-top: 0.25em;
     font-variant: small-caps;
     list-style-type: none;


christianlupus:
I rebased your PR to prevent the intermixture with the changes of seyfeb. Please adjust locally, so that the branches can be kept separate and better managed.

christianlupus (changes requested):
Generally speaking: You are introducing a new syntax of pre- and postfixing  heading with underlines. However, there is already in the RecipeIngredient Vue component the method `isHeader` that uses the prefix `## `.

By adding another syntax we are breaking existing recipes. Could you fix that, please?

PR #1998, opened 2023-12-15
feat: prevent yield calculation for ## as ingredient headline

j0hannesr0th:
## Topic and Scope

fixes https://github.com/nextcloud/cookbook/issues/1985

https://github.com/nextcloud/cookbook/assets/36242595/1ee962bc-bbb0-4d29-ac88-f787aa8e62a4

## Concerns/issues

none.

## Formal requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


christianlupus (approved):
LGTM

PR #2014, opened 2023-12-19
Fix/1408 more consistent UI timer

### LINKED ISSUE #1408: Make timer UI more consistent
I noticed some inconsistencies in the timer UI in the recipe view page:

- The label says `(H:MM)`, but this is not true once the timer is started. When the timer is running, it shows `H:MM:SS`. 
- The label in the recipe view page says `(H:MM)` whereas the label in the recipe edit page says `(hours:minutes)`.
- The seconds are only shown when the timer is running, which in itself is inconsistent.
- I think removing `(H:MM)` from the label (also from the recipe edit page) would make the UI cleaner. As an alternative, we could make it clear in the timer's value what each number represents.

In the recipe view page, the running timer shows `0:00:45` which does not match the label `(H:MM)`
image

The recipe edit page uses `(hours:minutes)`
image

I would like to propose removing `(H:MM)` from the labels and replacing `0:00:45` in the timer value with `0h 00m 00s`. This makes it clear what each value represents without the need for a legend. We can also do this in the recipe edit page. 

We could also replace the two different text boxes in the recipe edit page with a single box where the user enters `1h20m` for example. It could start with `00h00m`, and we could automatically move the cursor over after the two hour digits are typed by the user. We could even support writing arbitrary durations like `1d20h30m45s`. I've seen similar UX on other websites for durations, phone numbers, and things like that. I see pros and cons to this though.

### LINKED ISSUE #1747: Allow specifying seconds in timers
Hi,

I tried using Cookbook for coffee recipes and it's often required to specify time in seconds, while Cookbook allows setting only hours and minutes.

I think an additional field for seconds would be fine.

seyfeb:
## Topic and Scope

- Update timer display
  - Remove `H:MM` from label
  - Add units
  - Add support for durations with seconds
- Add possibility to define seconds for time periods in recipe editor 
- Make global i18n `n(...)` function available in JS

Closes #1408. Closes #1747 

Screenshot 2023-12-19 at 14 47 10

## Concerns/issues

_This currently only supports translations to left-to-right languages, i.e., first hours, then minutes, then seconds._

## Formal requirements

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -12,6 +12,7 @@ extends:
 
 globals:
     OC: readonly
+    n: readonly
     t: readonly
 
 rules:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ### Added
 - Toast with success/error message after trying to copy ingredients [#2040](https://github.com/nextcloud/cookbook/pull/2040) @seyfeb
+- Seconds can now be specified for recipe times [#2014](https://github.com/nextcloud/cookbook/pull/2014) @seyfeb
 
 ### Fixed
 - Prevent yield calculation for ## as ingredient headline [#1998](https://github.com/nextcloud/cookbook/pull/1998) @j0hannesr0th
+- Improved styling of times in recipe view [#2014](https://github.com/nextcloud/cookbook/pull/2014) @seyfeb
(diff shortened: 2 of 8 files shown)


christianlupus (approved):
Looks good. I have not verified every single line but it seems o work out well.

Issue #2002, opened 2023-12-17
Being prompted with a v0.10.5 update however its not in the changelog


krisgesling:
**Description**
Hi there, 

I was about to update the Cookbook app on my NC to a reported v0.10.5 and went to check the changelog but it doesn't seem that such a version exists in `CHANGELOG.md`. It also seems that the last version bump in Github Actions was [from 0.10.3 to 0.11.0-rc1](https://github.com/nextcloud/cookbook/commit/7d1b9cb629d650f788a79854024089e137a6bba1). 

I'm not familiar with the deployment/update workflow for NC apps but thought it worth mentioning. In the best case it's just been a manual push and there's no clear notes. Potentially it's actually updating to v0.10.3 and incorrectly reporting the version number. Worst case the h4x0r5 are coming for our cheesecakes via a malicious Cookbook update!!! :laughing: 

**Reproduction**
1. Install v0.10.2 of Cookbook
2. Go to see available App updates
3. Prompted to update to v0.10.5

**Expected behavior**
Only known versions of the Application should be presented. 
Changelog should have detail on any deployed version. 
Presumably the Github Actions (haven't looked) trigger a deployment based on the contents of `last_release`

**Versions**
Nextcloud server version: 27.1.3.2
Cookbook version: 0.10.2

christianlupus:
First, I can relax you a bit. The version 0.10.5 has been published by me, one of the admins of this repo.

I want to explain, what happened here: We had to update a dependency with the change of NC 28. This forced us to have a separate versioning scheme for pre-28 and 28. This was also announced in https://github.com/nextcloud/cookbook/discussions/1917. There is a [main-0.10.x branch](https://github.com/nextcloud/cookbook/tree/main-0.10.x) in this repository that holds the further development of the version scheme 0.10.

If you have a look at the changelog in the main-0.10.x branch, you will find the releases 0.10.3 to 0.10.5. Also, in the [tags of the repo](https://github.com/nextcloud/cookbook/tags) these versions are present.

You will not find released tarballs in this repo for newer releases. This is due to a security concern in the past. We had to outsource the release action into a separate repository outside the nextcloud organisation. So, the actual releases are located at https://github.com/christianlupus-nextcloud/cookbook-releases/releases.

There are thoughts of changing the changelog format by me and seyfeb but we have not yet decided on anything. Anyhow, the main effect would most probably still be present, I guess.

---

[...]

PR #2016, opened 2023-12-19
Fix/backport 2001 add missing translation

seyfeb:
## Topic and Scope

Backport of #2015 

## Concerns/issues

## Formal requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [Unreleased]
 
 ### Fixed
+- Prevent yield calculation for ## as ingredient headline [#1998](https://github.com/nextcloud/cookbook/pull/1998) @j0hannesr0th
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
christianlupus, line 1: I think the changes in this file are somehow intermixed with the changes of the master change log. Was this intentional?

seyfeb: Well, kind of, yes. Since we have not made any decisions how to deal with the CHANGELOG for backports/on the 0.10.x branch, I thought the easiest way would probably be to keep them in sync. However, they do not reflect the actual changes in the branch then, which would be kind of confusing. 

Keeping them in sync is certainly related to #2002 .

What we could do, is keeping in the CHANGELOG only changes for the respective branch and introduce additional "Release Notes" which contains the info about all releases in the master branch? 

Not sure what would be the best way to move forward.

christianlupus (approved):

CHANGELOG.md
christianlupus: Let's continue this discussion in #2021.

PR #2023, opened 2023-12-23
Feat/507 add citation field

### LINKED ISSUE #507: Additional field for source or citation
Hello!

I think it would be a good idea to have a general purpose _source_ field. There is the URL field, which I would use if I found the recipe on the Internets. But where to put the information, that the recipe is from _The Great Cooking Book, p. 362_ or _My best friend's Grandma_? Right now there is no suitable field for that.

Having a look at https://schema.org/Recipe, maybe the https://schema.org/citation field would fit.

seyfeb:
## Topic and Scope

Adds support for the `citation` property (https://schema.org/citation) to recipe editor and viewer.

Example with `citation: "Peter Pan"`

Screenshot 2023-12-23 at 18 10 20

Closes #507 

## Concerns/issues

Possible UI/UX option: Make citation text an anchor if `URL` property is set. This would make for a cleaner look (no URL is shown). Caveats: What if URL is set, but no `citation` field?

## Formal requirements

### DIFF
--- a/src/components/RecipeEdit.vue
+++ b/src/components/RecipeEdit.vue
@@ -17,6 +17,15 @@
                 :suggestion-options="allRecipeOptions"
             />
             <EditInputField
+                v-model="recipe['citation']"
+                :field-type="'text'"
+                :field-label="t('cookbook', 'Source')"
+                :placeholder="
+                    // TRANSLATORS Example (placeholder) name for a citation of the recipe's source
+                    t('cookbook', 'Grandma Betty')
+                "
+            />
+            <EditInputField
                 v-model="recipe['url']"
                 :field-type="'url'"
                 :field-label="t('cookbook', 'URL')"
@@ -201,6 +210,7 @@ const recipe = ref({
     id: 0,
     name: '',
     description: '',
+    citation: '',
     url: '',
     image: '',
     prepTime: '',
@@ -531,6 +541,7 @@ const initEmptyRecipe = () => {
         id: 0,
         name: null,
         description: '',
+        citation: '',
         url: '',
         image: '',
         prepTime: '',
--- a/src/components/RecipeView/RecipeView.vue
+++ b/src/components/RecipeView/RecipeView.vue
@@ -14,9 +14,18 @@
                 </div>
 
                 <div class="meta">
-                    <h2 class="heading">{{ $store.state.recipe.name }}</h2>
+                    <div class="heading">
+                        <h2 class="mb-0">{{ $store.state.recipe.name }}</h2>
+                        <p v-if="$store.state.recipe.citation">
+                            {{
+                                // TRANSLATORS Indicates citation/source of recipe. Ex. "by Grandma Betty"
+                                t('cookbook', 'by')
+                            }}
+                            <span>{{ $store.state.recipe.citation }}</span>
+                        </p>
+                    </div>
                     <div class="details">
-                        <div v-if="recipe.keywords.length">
+                        <div v-if="recipe.keywords.length" class="mb-3">
                             <ul v-if="recipe.keywords.length">
                                 <RecipeKeyword
                                     v-for="(keyword, idx) in recipe.keywords"
@@ -816,6 +825,12 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+.mb-0 {
+    margin-bottom: 0 !important;
+}
+.mb-3 {
+    margin-bottom: 0.75rem !important;
+}
 .wrapper {
     width: 100%;
 }
@@ -873,6 +888,7 @@ export default {
 
 .heading {
     margin-top: 12px;
+    margin-bottom: 1rem;
 }
 
 .dates {
--- a/src/components/RecipeView/RecipeView.vue
+++ b/src/components/RecipeView/RecipeView.vue
@@ -828,9 +828,11 @@ export default {
 .mb-0 {
     margin-bottom: 0 !important;
 }
+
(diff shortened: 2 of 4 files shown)


src/components/RecipeView/RecipeView.vue
```diff
@@ -14,9 +14,18 @@
...
                 <div class="meta">
-                    <h2 class="heading">{{ $store.state.recipe.name }}</h2>
+                    <div class="heading">
+                        <h2 class="mb-0">{{ $store.state.recipe.name }}</h2>
+                        <p v-if="$store.state.recipe.citation">
+                            {{
+                                // TRANSLATORS Indicates citation/source of recipe. Ex. "by Grandma Betty"
+                                t('cookbook', 'by')
+                            }}
+                            <span>{{ $store.state.recipe.citation }}</span>
```
christianlupus, line 24: This is not a good way to handle this wrt i10n. In other languages, it might not be in the pattern _static text_ plus name. So, this might be better delegated to the translators.

src/components/RecipeEdit.vue
```diff
@@ -16,6 +16,15 @@
                 :field-label="t('cookbook', 'Description')"
                 :suggestion-options="allRecipeOptions"
             />
+            <EditInputField
+                v-model="recipe['citation']"
```
christianlupus, line 20: I am not so sure about this usage of the `citation` field. According to the [schema.org standard](https://schema.org/citation), This field seems inappropriate according to the definition of this field:

> A citation or reference to another creative work, such as another publication, web page, scholarly article, etc.

This does not match the use case, especially with the _Grandma Betty_ placeholder content. How about using `author` or the `creator` fied.

I find it hard to quickly merge this PR now and soon have to create some sort of migration/fix that changes the code a posteriori.

christianlupus (changes requested):
I have some concerns. ne can be quickly fixed. The other needs some more discussion, I guess?

src/components/RecipeView/RecipeView.vue
seyfeb: Strictly speaking, you are right. I made this decision to allow different styles for the word "by" and the source string, which might even become an HTML element later on, like

```
<p>by <a :href="source-uri">Peter Pan</a></p>
```

But it would be better to translate the whole string including the citation variable, then HTML-escaping the citation, string-replacing the citation in the translated string with the html-escaped one (plus any additional styling, linking, etc) and using v-html in a span. Maybe there is a simpler method that I don't know of?

src/components/RecipeEdit.vue
seyfeb: The "not so sure" part is true for the other fields to. So thank you for initiating the discussion! 

According to schema.org the `author` and `creator` fields are the same 

> `creator`: The creator/author of this CreativeWork. This is the same as the Author property for CreativeWork.

However, I tend towards the creator being the one that creates the recipe in the cookbook, i.e., the Nextcloud user (analog to the fact that the `dateCreated` field is the non-user-configurable date the recipe was created, and not the date when grandma invented the recipe). This could become relevant when real sharing of cookbooks, recipes, collections, ... between users becomes a thing. It would allow filtering recipes by the user who entered the recipe in the cookbook app. As usual, we could store this info outside of the JSON ...

On the other hand, what I, as a creator, am doing when creating the recipe is citing "My grandma", who told me about the  list of ingredients and steps to take, but not necessarily the exact wording of each instruction, or the exact format for entering ingredients. I also might add some more comments on how grandma used to watch birds in the front yard while stirring the sauce but still cite her because she originally had the idea.

I do see the point though, that when I want to cite 

```
The Science of Cooking - Every Question Answered to Perfect Your Cooking,
 Stuart Farrimond, Dorling Kindersley Limited, 2017
 ``` 

the approach with "by XXX" falls a bit short and the `creator` (or `author`) field would make it much clearer what to expect from it, i.e., the "by XXX" syntax could be expected. 

One option would be to get rid of the word "by" and leave it up to the user to format the string to be shown using Markdown, which would also immediately allow linking. However my idea of moving the `url` property to a href of the citation to clean up the UI when viewing a recipe would then be off the table. I think, it's the usual consideration configurability vs. cleanness/ease of use.

Wrapping up, I see how both options have advantages and disadvantages. I'm looking forward for your opinion ;)

src/components/RecipeView/RecipeView.vue
christianlupus: How about something like `'by %s%s%s'` with the three parts `<a href="...">`, `$store.state.recipe.citation`, and `</a>`? In the first part, we would only have to insert the `href` value which can be handled manually in this case.

Theoretically, one could fall back to `by %s` and do the concatenation in JS directly.

src/components/RecipeEdit.vue
christianlupus: The general problem I see is the distinction in schema.org between a natural person/organization and a third (creative) work. The former is referencable by `author`/`creator`, while the latter is referenced by `citation`. In my understanding, these associations are mutually exclusive in some sort.

How about the following idea: We could have both fields (`citation` and `author`). One could be used as a source (in the sense of a reference), that would be a `citation`, while Grandma's best recipes could be declared using `author`. Would that work?

Of course, the UI must be thought through here as well. We would have 4 cases (no field, either field is set, or both fields are set). If none is set, we are at status quo. With one field set, your approach might work well (_by author_ or _from citation_ for example). With both present, we could think about it but I would then go with _by author_ and add a reference later on like the URL source currently.

PR #2059, opened 2024-01-14
Chore/add typescript support

### LINKED ISSUE #2053: Add Typescript support
Typescript has several benefits compared to the plain JavaScript we are currently using. Adding support for typescript and slowly migrating the app will hopefully result in better code maintainability (more structure, type checking, better documentation) and less bugs (type checking)

seyfeb:
## Topic and Scope

Added support for typescript

Closes #2053

## Concerns/issues

_None_

## Formal requirements


christianlupus (approved):
I do not see many changes except for stylistic ones (which do not count). 

Please check that it is working, eventually fix the comments you mentioned and you can merge this.

seyfeb:
Probably yes. I guess much of new frontend code could/should use typescript while existing code may be refactored into typescript. So if we want to continue to backport things the 0.10.x branch should have ts support, too.

PR #2099, opened 2024-01-26
fix: yield not set calculation error

### LINKED ISSUE #1986: Prevent recalculation algorithm if no yeild is given
The recalucaltion algorithm of the ingredients fails if the recipe yield is non-numeric or null. This will result in NaN entries while recipe viewing. This should be fixed.

j0hannesr0th:
Fixes #1986

Thanks for the video @christianlupus - I haven't knew that I can click "servings" to disable them. When trying to reproduce it I just haven't set it and it was "1" by default.

image

_Please write, what the PR is about and what it does or should do. This does not need to be very long._

none.

## Formal requirements


christianlupus (approved):
LGTM

PR #2225, opened 2024-03-31
Fix: remove client 'nook' from docs

shagn:
## Topic and Scope

The PR removes the client 'nook' from the clients page since it is not longer available. The link to the Play Store does not work and also with a search I could not find it. 

## Concerns/issues

No

## Formal requirements


christianlupus:
I will have to check this out.

A quick research found https://apkpure.net/de/nook-nextcloud-cookbook/org.capacitor.cookbook.app and https://apkpure.com/de/nook-nextcloud-cookbook/org.capacitor.cookbook.app/download.

Lets see if we can reach the developer.

christianlupus (approved):
Temporarily accepted until a clear statement from the developer comes to me. Until then we drop the Nook links

PR #2236, opened 2024-04-08
Update RecipeView.vue

philiprenich:
Fix grid layout causing large white space
- Allow rows to be auto sized
- Use `gap` instead of an extra column
- Adjust row & column as necessary

## Topic and Scope

These changes to the grid layout will remove the excess white space on the recipe view, especially for small screens.

## Concerns/issues

Due to Vue's scoped styles, this should not affect anything else.

## Formal requirements

    - CSS only change. But apologies regardless as I have not set up a dev env for Nextcloud yet.

### DIFF
--- a/.changelog/current/2236-fix-css-grid-layout.md
+++ b/.changelog/current/2236-fix-css-grid-layout.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Fix CSS grid layout causing whitespace issues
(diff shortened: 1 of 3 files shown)


christianlupus:
@seyfeb does this collide with your work currently?

christianlupus (approved):

PR #2291, opened 2024-05-04
Create new method for changelog handling

### LINKED ISSUE #2021: Fix the changelog process
In #2002 it became obvious that the changelog process is not optimal. we should handle the changelog better. The following requirements are present:

- The changelog in the `master` branch represents the complete history of all versions also on the `main-*` branches.
- The file `CHANGELOG.md` should _not_ be changed on each PR to prevent conflicts during rebasing and backports
- During the release process, the CHANGELOG must be updated automatically without manual intervention

Steps to carry this out:

_T.B.D._

Sketch of a solution:

One idea is to put the changelog entries for each PR into a dedicated folder `.changelog`. Each PR has its own file. The filename is always beginning with the PR name and a dash (e.g. `1234-`) to guarantee uniqueness. Such a PR changelog is build similar to a changelog file. For example, for PR #2015, the file `.changelog/2015-add-missing-translation` could look like
```
### Fixed
- Add missing translatable string for recipe-creation button in empty list view
```
Optionally, before the `### Fixed`, there could be a line `Authors: @seyfeb` (or multiple) to declare the authors responsible for the PR.

The CHANGELOG file itself is not modified during normal development. The PR changelogs are kept within the individual branches (`master`, `main-*`). There is one more (isolated) branch `changelogs` to be created. This branch _must_ have linear history and will not be used by normal devs. It contains no files except for a basic README (to explain its usage) and version changelogs (just read on).

During the release process, the changelog needs to be fixed. So, any yet unreleased entries must be collected and combined into a _bundle_. This can be done in a script (part of the release scripts should be possible). The unreleased PRs are in fact the pr-changelogs in `.changelog` of the corresponding `master` or `main-*` branch. So, the script will iterate these files and create an output that represents a version-changelog. It needs to collect all change types (`Fixed`, `Changed`, `Added`, ...) and combine the texts, augment by a link to the PR (number is known from the file name) and author information (either from `Authors` field in file or from author mail in git commit). The result is similar to e.g.
```
## 0.10.4 - 2023-12-05

### Fixed
- Make app compatible with PHP 7.4
  [#1931](https://github.com/nextcloud/cookbook/pull/1931) @christianlupus
- Remove constant se in RecipeView
  [#1942](https://github.com/nextcloud/cookbook/pull/1942) @j0hannesr0th

### Maintenance
- Add PHP lint checker to ensure valid (legacy) PHP syntax
  [#1931](https://github.com/nextcloud/cookbook/pull/1931) @christianlupus
- Add backport script to simplify development
  [#1935](https://github.com/nextcloud/cookbook/pull/1935) @christianlupus
```

This is a version changelog (in this case from 0.10.4). The script during the release process will use this output and commit a file `changelogs/v0.10.4` to the `changelog` branch. The filename is derived from the tag name of the version to create. So, it could also be `changelogs/v0.11.0-rc1`.

Depending if this is a prerelease or not, the folder `.changelog` in the actual `master` or `main-*` folder must be cleaned up. Additionally, the file `CHANGELOG.md` must be reconstructed from all snippets in the `changelog` branch. Additionally, the `CHANGELOG.md` on the `master` branch must be updated as well if working on a `main-*` branch.

---

Before starting to implement this, do you think this might work, @seyfeb?

christianlupus:
## Topic and Scope

Closes #2021

This will provide a way to simplify updates to the changelog. In fact the dev is only required to careate an individual file per PR.

## Concerns/issues

Some things to consider

- [x] Update the workflow script to look for the changlog snippet
- [x] Document the process for authors
- [x] Migrate the `main-0.10.x` branch
- [x] Update the deployment scripts

There is still the open question on how to handle the chronological ordering of the releases (0.10.6 can come after 0.11.0). Should this be respected in the changelog building process? How is this to be stored?

## Formal requirements

I did not change the code base itself, so no issues here.

### DIFF
--- a/.changelog/current/1989-auto-generated.md
+++ b/.changelog/current/1989-auto-generated.md
@@ -1,0 +1,3 @@
+### Documentation
+
+- Improve  structure of `README.md`
--- a/.changelog/current/1998-auto-generated.md
+++ b/.changelog/current/1998-auto-generated.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Prevent yield calculation for ## as ingredient headline
--- a/.changelog/current/2003-auto-generated.md
+++ b/.changelog/current/2003-auto-generated.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Prevent recalculation algorithm if no yield is given
--- a/.changelog/current/2014-auto-generated.md
+++ b/.changelog/current/2014-auto-generated.md
@@ -1,0 +1,7 @@
+### Added
+
+- Seconds can now be specified for recipe times
+
+### Fixed
+
+- Improved styling of times in recipe view
--- a/.changelog/current/2015-auto-generated.md
+++ b/.changelog/current/2015-auto-generated.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Add missing translatable string for recipe-creation button in empty list view
--- a/.changelog/current/2037-auto-generated.md
+++ b/.changelog/current/2037-auto-generated.md
@@ -1,0 +1,3 @@
+### Added
+
+- New filter UI in recipe lists
--- a/.changelog/current/2040-auto-generated.md
+++ b/.changelog/current/2040-auto-generated.md
@@ -1,0 +1,3 @@
+### Added
+
+- Toast with success/error message after trying to copy ingredients
--- a/.changelog/current/2059-auto-generated.md
+++ b/.changelog/current/2059-auto-generated.md
@@ -1,0 +1,3 @@
+### Maintenance
+
+- Add Typescript support
--- a/.changelog/current/2099-auto-generated.md
+++ b/.changelog/current/2099-auto-generated.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Fix yield not set calculation error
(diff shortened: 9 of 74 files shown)


christianlupus:
Question for you, @seyfeb:
Do you think it is a required/good practice to create a common CHANGELOG (i.e. the CHANGELOG.md in the `master` branch) for **all** releases created (also the ones residing in the other branches like the `main-0.10.x`)? Currently, the CHANGELOG is distributed over the branches. Would it bring a significant benefit to have a single CHANGELOG file in the `master` branch?

seyfeb:
> Question for you, @seyfeb: Do you think it is a required/good practice to create a common CHANGELOG (i.e. the CHANGELOG.md in the `master` branch) for **all** releases created (also the ones residing in the other branches like the `main-0.10.x`)? Currently, the CHANGELOG is distributed over the branches. Would it bring a significant benefit to have a single CHANGELOG file in the `master` branch?

I don't think it would provide much benefit to put all the changes of different branches in a CHANGELOG file on the master branches. Combining the changes from all branches into a single changelog file on the main branch might even lead to confusion since it does not actually contain changes on the `master` branch as one would expect.

christianlupus:
This PR needs a companion PR (for the `main-0.10.x` branch) that I need to prepare before processing further.

PR #2343, opened 2024-05-30
Use underscores to prevent trailing dot in folder names

### LINKED ISSUE #2332: Long folder names create sync error
**Description**
#1620 Introduced the shortening of folder names with the name ending with `...`.  Sadly, this causes the windows sync client to throw errors because a file name must not end with a `.`

**Reproduction**
1. Create recipe with long name that will be shortened
2. Activate syncing the recipes folder on a windows sync client (with NTFS I think)

**Expected behavior**
No sync errors.
Maybe use another way to show that the name has been shortened, like appending `-short` for example.

**Actual behavior**
Sync errors

**Screenshots**
Error Message:
grafik
`File names ending with a dot are not supported by this file system.`

**Browser**
Firefox 126.0 (64-Bit)

**Versions**
Nextcloud server version: 27.1.9
Cookbook version: 0.10.5
Database system: MySQL

christianlupus:
## Topic and Scope

Closes #2332

## Concerns/issues

There could be name clashes if the first 97 chars of two recipes are equal.

## Formal requirements

### DIFF
--- a/.changelog/current/2343-no-dot-at-end-of-file-name.md
+++ b/.changelog/current/2343-no-dot-at-end-of-file-name.md
@@ -1,0 +1,3 @@
+# Fixed
+
+- Prevent dot at end of file/folder name when abbreviating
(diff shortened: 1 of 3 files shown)


christianlupus:
In case there are already some recipes with the trailing dot, the user has to open the recipe and just store it again. That will move it to the correct location without a trailing dot.

PR #2385, opened 2024-06-23
Fix #2384: Fix import of WP Recipe Maker recipes

### LINKED ISSUE #2384: Add importer for WP Recipe Maker (Wordpress Plugin)
**Failing website:** https://biancazapatka.com/de/vegane-koettbullar-schwedische-hackbaellchen/

**Checking if valid metadata are present:** Yes, I check the source code of the website and found metadata.

**Cookbook version:** ? (I assume latest)

**Problem description (if applicable):**
The page cannot be imported. The error message is: "No parser found for the given import."

The website seems to use this WordPress Plugin: https://wordpress.org/plugins/wp-recipe-maker/

nathanielhourt:
- **Fix recipe import from some WordPress sites**
- **Add test exercising new parser logic**

## Topic and Scope

Many recipe websites are built with WordPress Recipe Maker with Yoast SEO exposing the recipe as a parseable
schema. This combination presented a schema which the recipe extractor did not handle properly.

This PReq makes a small modification to the parser logic to support this new schema format, and adds a test
exercising the new functionality.

I have found several sites that previously failed to parse, and now parse successfully. I believe this update
resolves the issue with most WordPress Recipe Maker websites, and it definitely resolves the parsing failure
on the site referenced in #2384.

## Concerns/issues

The tests do not detect any regressions from this change, and I do not expect the change to be a significant
risk of causing regressions.

## Formal requirements


christianlupus (approved):
Looks good. Hopefully this will not break any other pages but we can fix that later on.

PR #2391, opened 2024-06-25
Remove vulnerable ws dev dependency

christianlupus:
This is just a quick audit fix of a vulnerability. The problem is not in the live code but just the dev environment, so no big deal anyways.


christianlupus:
-approve

PR #2416, opened 2024-07-11
Alphabetize Category and Keyword Filters

nathanielhourt:
## Topic and Scope

In using the Cookbook app, I want to organize my recipes by keywords (often, ingredients) that I find relevant
when searching for recipes to make. I use the Keywords feature for this; however, when filtering by the
keywords, the drop-down list is unsorted by any metric I can comprehend (it's probably sorted by hash value) so
I have difficulty finding the one I'm looking for.

In this PReq, I have alphabetized the Category and Keyword drop-down lists in the recipe lists.

## Concerns/issues

It passes tests, and it LGTM in deployment. The change is small; I don't think it should break anything.

## Formal requirements

- [x] I created tests for newly added PHP code (no PHP changes were made)
- [x] I updated the OpenAPI specs and added an entry to the API changelog (API was not modified)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3922,24 +3922,6 @@
       "resolved": "https://registry.npmjs.org/@types/escape-html/-/escape-html-1.0.4.tgz",
       "integrity": "sha512-qZ72SFTgUAZ5a7Tj6kf2SHLetiH5S6f8G5frB2SPQ3EyF02kxdyBFf4Tz4banE3xCgGnKgWLt//a6VuYHKYJTg=="
     },
-    "node_modules/@types/eslint": {
-      "version": "8.56.10",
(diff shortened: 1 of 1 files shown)


christianlupus:
@seyfeb again the question: how to incorporate with the rework?

seyfeb:
I think it is sensible to have them sorted alphabetically. We need to make sure that does changes carry over to the rework. I'll try to remember that :D

christianlupus (approved):

christianlupus:
This is not ideal as there might be keyword/categories that are only differencing in case (e.g. `sweet` vs `Sweet`). These will only appear once in the list now (:+1:) but it is not possible to get all recipes of a certain writing (only `sweet`) for manual fixing. I see this a feature for the future with the new UI

PR #2421, opened 2024-07-13
Handle fractional ingredients (Fixes #2004)

### LINKED ISSUE #2004: In recipes, fractions are usually used for the non-integer quantity, e.g. 1/2 tsp. The specification 0.5 tsp is rather uncommon.
In recipes, fractions are usually used for the non-integer quantity, e.g. 1/2 tsp. The specification 0.5 tsp is rather uncommon.
It would be nice if the automatic change from the input (e.g. 1/3) to 0.33 could be prevented by configuration.

_Ursprünglich gepostet von @cloud2018 in https://github.com/nextcloud/cookbook/issues/1987#issuecomment-1858842533_

Weissnix4711:
## Topic and Scope

My solution to #2004. This handles 16ths, else falls back to decimal. I think this is a nice balance between using some fractions, without ending up with stupid denominators.

## Concerns/issues

Will only work if the initial ingredient is given as a fraction. If it's a decimal, or just a whole number, will still remain decimal. Maybe it would make sense to convert everything to fractions where possible, but then maybe not. Perhaps being able to change the behavior in settings would be useful, idk. Also, by design does not do 1/3, 1/5, etc. I don't mind this, in fact I would prefer this way, but I guess in the end it's not up to me.

## Formal requirements

### DIFF
--- a/.changelog/current/2421-fix-fraction-ingrediants.md
+++ b/.changelog/current/2421-fix-fraction-ingrediants.md
@@ -1,0 +1,4 @@
+# Fixes
+
+- Fraction calculation keeps fractions if possible
+
--- a/src/js/yieldCalculator.js
+++ b/src/js/yieldCalculator.js
@@ -63,7 +63,17 @@ function recalculateIngredients(ingredients, currentYield, originalYield) {
 
             const decimalAmount = wholeNumberPart + numerator / denominator;
             let newAmount = (decimalAmount / originalYield) * currentYield;
-            newAmount = newAmount.toFixed(2).replace(/[.]00$/, '');
+            const newWholeNumberPart = parseInt(newAmount, 10);
+            let newNumerator = (newAmount - newWholeNumberPart) * 16;
+            if (Number.isInteger(newNumerator)) {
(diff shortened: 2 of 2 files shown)


seyfeb:
Thank you for your contribution! Smaller than 1/16 probably does not make any sense indeed. 1/3 and 1/5 are debatable, though. However, as a starting point, having any conversion is better than none. In future iterations there could even be a toggle in the recipe view or some options in the configuration but that is not required immediately.

I did not check the details yet, but one thing that comes to my mind is, there are ASCII/Unicode symbols for fractions like ¾. I think they should be handled, too. Maybe not all of them, but the most common ones?

A quick Google search turned up this reference:

https://www.compart.com/de/unicode/decomposition/%3Cfraction%3E

Weissnix4711:
Done. Should now parse unicode fractions, but will always display the output as 'normal' text (ie. using digits and a slash). I guess it would be possible to display the output with unicode fractions too, but at least for me they don't always display too nice in the browser.

christianlupus (approved):
Although there is a small bug there, this is better than before.

PR #2507, opened 2024-09-18
Migrate Webpack to Vite build system

christianlupus:
## Topic and Scope

This will migrate to Vite packer finally

## Concerns/issues

This is not finally tested :wink: 

## Formal requirements


christianlupus:
@taye, I just saw, why I knew a bit about the Webpack -> Vite migration :wink:

I have completely forgotten about this one. We can start from scratch or rebase/fix this one **if** we (you) want to go with Vite.

taye:
@christianlupus Nice :D. I'll start from here

PR #2622, opened 2024-12-15
Allow use of browserless server

mrleadhead:
## Topic and Scope

I wanted to import recipes from gousto.co.uk however they require JavaScript to load the schema json. To achieve this I added a setting for the browserless address and adjusted the HTML download so that if the setting is set it will proxy the html download via browserless. https://github.com/browserless/browserless#hosting-providers

Open to making this more general allowing other services to be used. Browserless seemed the simplest tool currently.   

## Concerns/issues

There is an issue where the browserless address field in the settings page does not populate, I've reached the end of my vue and php skills I'm sure it's something simple. Feel free to make as many changes as you like my php skill is not great.  

## Formal requirements

- [ X] I did check that the app can still be opened and does not throw any browser logs
- [ X] I updated the OpenAPI specs and added an entry to the API changelog (check if API was not modified)
- [ -] I notified the matrix channel if I introduced an API change

### DIFF
--- a/lib/Helper/Filter/JSON/RecipeIdTypeFilter.php
+++ b/lib/Helper/Filter/JSON/RecipeIdTypeFilter.php
@@ -8,11 +8,15 @@ namespace OCA\Cookbook\Helper\Filter\JSON;
  * The id should be a string and no integer.
  */
 class RecipeIdTypeFilter extends AbstractJSONFilter {
-	public function apply(array &$json): bool {
-		$copy = $json;
-		if (array_key_exists('id', $json)) {
-			$json['id'] = strval($json['id']);
-		}
+    public function apply(array &$json): bool {
+        $copy = $json;
+        if (array_key_exists('id', $json)) {
+            $json['id'] = strval($json['id']);
+        }
+        // Check and fix `id` under `author`
+        if (array_key_exists('author', $json) && is_array($json['author']) && array_key_exists('id', $json['author'])) {
+            $json['id'] = strval($json['author']['id']);
+        }
 
 		return $json !== $copy;
 	}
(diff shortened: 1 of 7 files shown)


lib/Helper/Filter/JSON/RecipeIdTypeFilter.php
```diff
@@ -8,11 +8,15 @@
...
-		}
+    public function apply(array &$json): bool {
+        $copy = $json;
+        if (array_key_exists('id', $json)) {
+            $json['id'] = strval($json['id']);
+        }
+        // Check and fix `id` under `author`
+        if (array_key_exists('author', $json) && is_array($json['author']) && array_key_exists('id', $json['author'])) {
+            $json['id'] = strval($json['author']['id']);
+        }
```
mrleadhead, line 19: Gousto put the ID under author to be difficult

SnowyLeopard:
Would love to get this merged, what's needed to get this moving?

christianlupus:
I quickly skimmed over the PR as I closed all open PRs for the maintenance release 0.11.3. I saw, that I would like to look closer on this.

Just one point I directly saw was the fact that the browserless approach was not using curl but file_get_contents. This is generally fine but caused some trouble in the past (we used file_get_contents there for all downloads) with broken encodings (especially Asian and kyrillic languages).

Also, I am not sure about a good architecture here. I am thinking about using polymorphism here to reuse the curl setup at best. But this just a quick glance at the code. 

I would like to give it a good glance and see to a good integration of it all, if you do not mind. I find this interesting as it might solve some issues we have right now. However I was not aware of this and needed to get known to browserless.

christianlupus:
What I found that needs to be done: 

- [ ] make the code blend in nicely
- [ ] use curl to avoid encoding issues
- [ ] add a section to the admin documentation what this is about
- [ ] add tests for the integration in PHP (the checkmark was set wrongly, there are no updated tests) 
- [ ] the CI needs updating to create a browserless instance for the tests
- [ ] eventually, secrets need to be handled. We have to check this in detail.
- [ ] dev documentation might need updating

Issue #1706, opened 2023-05-29
Remove psalm errors


christianlupus:
Currently, there are quite done psalm issues detected and hidden by the baseline. These issues should be fixed by using correct annotations. 

It might be questionable is the baseline should be updated upon pr merging automatically.

christianlupus:
See also #990

provokateurin:
If you don't mind I can also fix the rest which is only stubs AFAIK.

christianlupus:
I think 2 classes (`oc_image` and `OC\Hooks\Emitter`) cause trouble. So, you suggest to copy the required class methods into files in the cookbook source code without actual content and located such that these are not interfering with the DI algorithm?

provokateurin:
Yeah basically that. I've done it for a few apps now, so it's relatively easy for me. Besides those classes, often doctrine and symfony classes are needed as well because they are used somewhere.

PR #2628, opened 2025-01-04
Fix a lot of psalm issues

provokateurin:
## Topic and Scope

https://github.com/nextcloud/cookbook/issues/1706

All the issues with JSONResponse will go away when https://github.com/nextcloud/server/pull/50035 is merged and OCP is updated here.

I think after that all remaining issues are related to missing stubs, which I can also make a PR for after this one is merged.

## Formal requirements

### DIFF
--- a/.changelog/current/2628-fix-php-issues.md
+++ b/.changelog/current/2628-fix-php-issues.md
@@ -1,0 +1,4 @@
+# Maintenance
+
+- Clean PHP code by removing code smells and redundant code
+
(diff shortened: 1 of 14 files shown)


provokateurin:
Fixes from https://github.com/nextcloud/server/pull/50035 are now included, only stubs are missing to have an completely empty baseline.

christianlupus (approved):
Looks good. Thanks for fixing these

provokateurin:
I wasn't sure if I was supposed to update the changelog, since these changes are nothing that will impact any user.

christianlupus:
That is true, but otherwise, the changelog guard will bite you :stuck_out_tongue:. And I put it in the Maintenance category, that is okay, I think.

Issue #1209, opened 2022-09-15
Import from frischgekocht.billa.at fails with non-descriptive error.


MacLemon:
**Failing website:**

**Checking if valid metadata are present:**
- [x] Yes, I check the source code of the website and found metadata.
  - Source contains `<article class='recipe is-new' itemscope itemtype="http://schema.org/Recipe">`

**Cookbook version:**
- Cookbook  Version 0.9.15

**Problem description (if applicable):**

# Summary:
Importing from https://frischgekocht.billa.at/ fails with error.

## Steps to Reproduce:
1. Navigate to Cookbook on your instance. `/apps/cookbook/`
1. Focus the “Download recipe from URL” field
1. Paste this recipe URL: https://frischgekocht.billa.at/rezept/kuerbis-erdaepfel-gulasch-mit-debreziner-BI-28065
1. Error is displayed

## Expected Results:
Recipe should be imported into my collection.

## Actual Results:
1. Nextcloud/Cookbook reports an error dialog with “The server reported an error. Please check.”.

The browser Log shows this info:

```
POST https://blackmoor.maclemon.at/apps/cookbook/webapp/import 500
(anonymous) @ xhr.js:220
e.exports @ xhr.js:16
e.exports @ dispatchRequest.js:56
d.request @ Axios.js:109
(anonymous) @ bind.js:9
import @ api-interface.js:48
downloadRecipe @ AppNavi.vue:271
Ft @ vue.runtime.esm.js:2987
n @ vue.runtime.esm.js:1785
Ft @ vue.runtime.esm.js:2987
e.$emit @ vue.runtime.esm.js:3685
onSubmit @ ActionInput.js:2
submit @ ActionInput.js:2
Ft @ vue.runtime.esm.js:2987
n @ vue.runtime.esm.js:1785
Qr.a._wrapper @ vue.runtime.esm.js:7428
[...]

PR #2711, opened 2025-03-24
Fix Microdata parser not understanding itemscope properly

### LINKED ISSUE #1617: Cannot parse recipe from swrfernsehen.de/kaffee-oder-tee
**Failing website:** https://www.swrfernsehen.de/kaffee-oder-tee/rezepte/eier-curry-100.html

**Checking if valid metadata are present:**

[x] Yes, I check the source code of the website and found metadata.
* `<meta name="itemtype" content="http://schema.org/Recipe">`

**Cookbook version:**
0.10.2

**Problem description (if applicable):**

The page cannot be imported. The error message is:
```
Exception: OCA\Cookbook\Helper\TextCleanupHelper::cleanUp(): 
Argument #1 ($str) must be of type ?string, array given, called in 
/var/www/html/custom_apps/cookbook/lib/Helper/Filter/JSON/RecipeNameFilter.php on line 26 in file '/var/www/html/custom_apps/cookbook/lib/Helper/TextCleanupHelper.php' line 14

0. /var/www/html/lib/private/AppFramework/App.php - line 172:
OC\AppFramework\Http\Dispatcher->dispatch([ "OCA\\Cook ... "], "import")
1. /var/www/html/lib/private/Route/Router.php - line 298:
OC\AppFramework\App::main("OCA\\Cookbo ... r", "import", [ "OC\\AppFr ... "], [ "cookbook.recipe.import"])
2. /var/www/html/lib/base.php - line 1047:
OC\Route\Router->match("/apps/cookbook/webapp/import")
3. /var/www/html/index.php - line 36:
OC::handleRequest()

Caused by TypeError: OCA\Cookbook\Helper\TextCleanupHelper::cleanUp(): 
Argument #1 ($str) must be of type ?string, array given, called in 
/var/www/html/custom_apps/cookbook/lib/Helper/Filter/JSON/RecipeNameFilter.php on line 26

0. /var/www/html/custom_apps/cookbook/lib/Helper/Filter/JSON/RecipeNameFilter.php - line 26:
OCA\Cookbook\Helper\TextCleanupHelper->cleanUp([ "SWR","SWR ... "], true, true)
1. /var/www/html/custom_apps/cookbook/lib/Helper/Filter/JSONFilter.php - line 63:
OCA\Cookbook\Helper\Filter\JSON\RecipeNameFilter->apply([ "Recipe"," ... ]])
2. /var/www/html/custom_apps/cookbook/lib/Service/RecipeService.php - line 177:
OCA\Cookbook\Helper\Filter\JSONFilter->apply([ "Recipe"," ... ]])
3. /var/www/html/custom_apps/cookbook/lib/Service/RecipeService.php - line 324:
OCA\Cookbook\Service\RecipeService->checkRecipe([ "Recipe"," ... ]])
4. /var/www/html/custom_apps/cookbook/lib/Controller/Implementation/RecipeImplementation.php - line 222:
OCA\Cookbook\Service\RecipeService->downloadRecipe("https://www ... l")
5. /var/www/html/custom_apps/cookbook/lib/Controller/RecipeController.php - line 85:
OCA\Cookbook\Controller\Implementation\RecipeImplementation->import()
6. /var/www/html/lib/private/AppFramework/Http/Dispatcher.php - line 225:
OCA\Cookbook\Controller\RecipeController->import()
7. /var/www/html/lib/private/AppFramework/Http/Dispatcher.php - line 133:
OC\AppFramework\Http\Dispatcher->executeController([ "OCA\\Cook ... "], "import")
8. /var/www/html/lib/private/AppFramework/App.php - line 172:
OC\AppFramework\Http\Dispatcher->dispatch([ "OCA\\Cook ... "], "import")
9. /var/www/html/lib/private/Route/Router.php - line 298:
OC\AppFramework\App::main("OCA\\Cookbo ... r", "import", [ "OC\\AppFr ... "], [ "cookbook.recipe.import"])
10. /var/www/html/lib/base.php - line 1047:
OC\Route\Router->match("/apps/cookbook/webapp/import")
11. /var/www/html/index.php - line 36:
OC::handleRequest()
```

When importing, the recipe gets disturbed, because the page contains multiple meta data elements with `itemprop="name"`. Microdata strcure in tha html seems correct as other "name" properties are given within an "itemscope" (e.g. Organisation, Person, BreadcrumbList). When I parse the HTML with the page https://foolip.org/microdatajs/live/ the data is shown correctly.

krzys-h:
## Topic and Scope

This makes the simple Microdata parser properly handle nested objects. Here is a minimal repro example of the bug it fixes:
```html
<div itemscope itemtype="https://schema.org/Recipe">
  <span itemprop="name">Sample recipe name</span>
  <div itemprop="author" itemscope itemtype="https://schema.org/Person">
    <span itemprop="name">Sample person</span>
  </div>
</div>
```

Previously, the import would crash because both "name" props would be interpreted as the name of the recipe. Now, itemprops nested inside child itemscopes are ignored.

Closes #1617

## Concerns/issues

This PR should be extended with unit tests, but I don't have the entire dev environment set up to do it right now. I mostly just wanted to bodge a fix for personal use because the bug bothered me. I only tested that the three URLs mentioned in #1617 work fine now.

## Formal requirements


lib/Helper/HTMLParser/HttpMicrodataParser.php
```diff
@@ -208,7 +208,7 @@ private function searchSimpleProperties(DOMNode $recipeNode, string $property):
 	 * @return DOMNodeList A list of all found child nodes with the given property
 	 */
 	private function searchChildEntries(DOMNode $recipeNode, string $prop): DOMNodeList {
-		return $this->xpath->query(".//*[@itemprop='$prop']", $recipeNode);
+		return $this->xpath->query(".//*[@itemprop='$prop' and count(ancestor::*[@itemscope]) = 1]", $recipeNode);
```
christianlupus, line 211: I am unsure about this XPath stuff right now. Especially the hard-coded `1` makes me nervous.

Also one test case, that I can think of (and that might cause problems) but I have not found a concrete example of: Let the top-most element be a non-recipe. Like a [WebPage](https://schema.org/WebPage). The WebPage is [about](https://schema.org/about) a Recipe. That way, the Recipe is no longer the top-most element. How will `ancestor::*[@itemscope]` then behave? Will it detect the recipe's props or the Webpages?

Maybe you could even create a test case (similar to the PR introduction) and add similarly to the test suite?

christianlupus (changes requested):
I added the HTML pages from the linked issue as test cases and also the test case from you. I tried to extract useful output information mainly based on the current implementation (so, this might be wrong, please check the added JSON files).

All in all, your tests made quite some tests fail. I do not see any obvious errors in the test data. Can you please elaborate what is the problem?

Adding a changelog will at least make this test succeed easily.

lib/Helper/HTMLParser/HttpMicrodataParser.php
krzys-h: Most of the failing test cases fail because... they are not actually valid Microdata. If you run them through https://validator.schema.org/ you will see that it can't see any elements either, same goes for https://foolip.org/microdatajs/live/. The spec says:

> The `itemtype` attribute can only be specified on elements which have an `itemscope` attribute specified.
> https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/itemtype

But these files specify an `itemtype` without an `itemscope`. I added this requirement to make sure the lookup for the parent `itemscope` works correctly. We can try to relax this requirement, but if even the official validator can't see them, then I don't expect this happens often in the real world. How would you like me to proceed here, add the missing `itemscope` to the test cases, or make the parser more lenient?

This is the case for: `caseB.html`, `caseC.html`, `caseD.html`, `caseE.html`, `caseImageAttribute.html`, `caseImageContent.html`, `caseIngredient.html` and `caseInstruction.html`.

---

The only actually failing test case is `caseFix1209.html`, and it seems to be exactly the issue with the parent `WebPage` itemscope you mentioned. I'm not really good with XPath so I wrongly assumed the `ancestor` query would be limited to the subtree of the node specified through `$recipeNode`, but this it not the case. Sadly I can't find a way around this without using XPath 2.0, which is not supported by `DOMXPath`. The easiest way I can see to fix this is to create a child `DOMDocument` containing only the recipe using `DOMDocument::importNode`, create a separate `DOMXPath` instance for it, and keep the current query that checks if we are not deeper than 1 itemscope from the root. Would you like me to do that, or perhaps we should replace the XPath queries with manual iteration over the nodes in PHP?

christianlupus: OK, I think, I fixed the cases you mentioned (B through E, Image*, as well as Ingredient and Instruction). I think I fixed these in the tests.

Regarding the test Fix1209, this was added for #1209 and its fix in #1220. I am a bit more concerned as this might break the existing code somewhat. For the given test case, this is maybe even a _good break_ (as it actually separates the instructions at least partially). However, it drops the first instruction without notice.

I marked it as incomplete for the meantime. If you see any good way to fix this or to work around it, feel free to give me a hint (or better open another PR :wink: )

---

Regarding the new parser routines, please do not put too much effort into it. There is a plan to rework the complete parser routines (also for the JSON+LD and other parsers). So, this would be wasted time. You are welcome to help with the rework if you want to. The point is to put the parsers on a common base.

christianlupus (approved):
LGTM

The missing changes are longer term and this should only be a quick fix PR. So, let's do smaller steps.

Issue #2826, opened 2025-08-18
Acquire wakelock when showing recipes


provokateurin:
**Is your feature request related to a problem? Please describe.**

On my phone I use the cookbook PWA because none of the available Android apps really satisfy me. For most things I'm really happy how it works, the only annoying thing (besides an obvious lack of offline support which I don't really consider in scope for a Nextcloud web app) is the lack of a wakelock when viewing a recipe. I usually put my phone on the table while cooking and have a very low timeout for the screen lock. In my other phone usage this is not a problem and I do it for security/privacy reasons, but while cooking in my own kitchen it is a bit annoying that I constantly need to unlock it.

**Describe the solution you'd like**

While viewing a recipe the frontend should acquire a screen wake lock with the dedicated API: https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API

---

This is probably be really easy to implement, so I might give it a go myself if there is time for it.

PR #2828, opened 2025-08-21
feat(RecipeView): Acquire screen wake lock

provokateurin:
## Topic and Scope

Closes https://github.com/nextcloud/cookbook/issues/2826

While viewing a recipe the screen of the device will no longer lock automatically, making it easier to follow a recipe while cooking in parallel.

## Concerns/issues

None that I can think of.

## Formal requirements


christianlupus (approved):
I tried to verify the changes but did not succeed. Unfortunately, I was not able to use a secure context.

PR #2881, opened 2025-10-13
Update de.json

Gugiman:
Change translation "fiber" -> "Faser" to the better known term "Ballaststoffe" 

At least in this context, this translation makes more sense

### DIFF
--- a/l10n/de.json
+++ b/l10n/de.json
@@ -137,7 +137,7 @@
     "Sugar" : "Zucker",
     "Carbohydrate" : "Kohlenhydrate",
     "Cholesterol" : "Cholesterin",
-    "Fiber" : "Faser",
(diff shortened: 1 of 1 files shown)


christianlupus:
Unfortunately, we cannot change this here. It must be fixed in transifex, otherwise it will be overwritten at the nightly sync.

christianlupus:
@rakekniven, you might know more than me :wink:. I saw, the [translation](https://app.transifex.com/nextcloud/nextcloud/translate/#de/cookbook/581826280?q=text%3Afiber) is already reviewed and it seems no changes are possible currently. Is it possible to _un-review_?

rakekniven:
> Is it possible to un-review?

yes, every reviewer can do this. 

Adapted informal german version

PR #2990, opened 2026-02-02
Maint/test against nc 33

christianlupus:
## Topic and Scope

This is a quick update of the test infrastructure to run any tests against the current stable 33 branch of NC server.

## Concerns/issues

I had to mark the app as compatible with NC33 to make the test run. Thus, this should **not** be merged until support with NC 33 is actually established.

## Formal requirements


christianlupus:
This is just a dummy PR that is not actually needed. It was required to test some issue. I close this PR therefore.

Issue #2180, opened 2024-03-08
New "Nutrition Information" option "Dietary Restriction"


nicodemusjls:
I'd like a new option added to the drop-down list of "Nutrition Information". Please add "Dietary Restriction Compliance" or similar as a way of adding a list of dietary restrictions that the recipe is complaint with. In use, we would select this new option then manually type the list of allergens it doesn't have or diets it conforms to.  I know we can use the "categories" section to get a similar function, but I think this is better kept out of the other types of categories AND it should print along side the Nutrition Information section.

Example:
Dietary Restriction Compliance- peanut allergy, nightshades allergy, vegan, vegetarian

Further, it would be awesome if the text box also worked just like the categories item works, where we can select from existing or add new phrases. However, as a quick, partial fix, we could just have the option added with a standard text box.

seyfeb:
Thank you for your suggestion. We follow the https://schema.org/Recipe for structuring the information. The `NutritionInformation` object [does not contain](https://schema.org/NutritionInformation) dietary information but only the nutritional values. 

If I understand you correctly your requested feature might be achievable using the [suitableForDiet](https://schema.org/suitableForDiet) field.

jessebot:
Came here when searching the issues for [`suitableForDiet`](https://schema.org/suitableForDiet), which I'd like to see added. It uses the following values from [`RestrictedDiet`](https://schema.org/RestrictedDiet):

- [DiabeticDiet](https://schema.org/DiabeticDiet)
- [GlutenFreeDiet](https://schema.org/GlutenFreeDiet)
- [HalalDiet](https://schema.org/HalalDiet)
- [HinduDiet](https://schema.org/HinduDiet)
- [KosherDiet](https://schema.org/KosherDiet)
- [LowCalorieDiet](https://schema.org/LowCalorieDiet)
- [LowFatDiet](https://schema.org/LowFatDiet)
- [LowLactoseDiet](https://schema.org/LowLactoseDiet)
- [LowSaltDiet](https://schema.org/LowSaltDiet)
- [VeganDiet](https://schema.org/VeganDiet)
- [VegetarianDiet](https://schema.org/VegetarianDiet)

@nicodemusjls I think this is probably what you need anyway, so this could issue title could be updated to say "Support `suitableForDiet`".

nicodemusjls:
[...]

PR #3020, opened 2026-03-01
feat: add suitableForDiet field from schema.org RestrictedDiet vocabulary

samtuckerdavis:
## Summary

Adds support for the `suitableForDiet` property from the [schema.org RestrictedDiet vocabulary](https://schema.org/RestrictedDiet), closing #2180.

The schema.org `Recipe` type has supported `suitableForDiet` for several years. Websites that export recipes using JSON-LD (and some scrapers) already include this field. Without a normalisation filter, those values pass through unmodified when they happen to be well-formed, or get silently dropped when they are not.

## Changes

**`lib/Helper/Filter/JSON/FixSuitableForDietFilter.php`** (new)
A new `AbstractJSONFilter` subclass that mirrors the established pattern of `FixKeywordsFilter`:
- Initialises the field to `''` when absent
- Normalises arrays (produced by some scrapers) into comma-separated strings
- Strips HTML tags, trims whitespace, deduplicates entries
- Logs a warning and resets to `''` when the value is an unexpected type

**`lib/Helper/Filter/JSON/JSONFilter.php`**
Registers `FixSuitableForDietFilter` in the filter pipeline after `FixNutritionFilter`.

**`tests/Unit/Helper/Filter/JSON/FixSuitableForDietFilterTest.php`** (new)
Unit tests following the structure of `FixKeywordsFilterTest`, covering missing field, invalid type, string input, array input, whitespace/HTML normalisation, and deduplication.

## Notes

- No database migration is required — the field is stored in the recipe JSON file alongside all other schema.org properties.
- The field is passed through transparently by the existing JSON read/write paths once the filter ensures it is present and normalised.
- Filter/output for the API response layer (e.g. `RecipeJSONOutputFilter`) can be extended in a follow-up if the field should be guaranteed present in API responses, consistent with the `nutrition` field handling.

### DIFF
--- a/lib/Helper/Filter/JSON/FixSuitableForDietFilter.php
+++ b/lib/Helper/Filter/JSON/FixSuitableForDietFilter.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace OCA\Cookbook\Helper\Filter\JSON;
+
+use OCA\Cookbook\Helper\TextCleanupHelper;
+use OCP\IL10N;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Fix the suitableForDiet field of a recipe.
+ *
+ * suitableForDiet is a schema.org RestrictedDiet property on the Recipe type.
+ * It is stored as a comma-separated list of diet identifiers (e.g. "VeganDiet,VegetarianDiet").
+ *
+ * If the field is absent, it is initialised to an empty string.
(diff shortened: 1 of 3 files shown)


samtuckerdavis:
Closing this PR — on reflection, this change may not be the right fit for this project. Thanks for your time!

PR #3084, opened 2026-04-05
Increasing compatibility with JsonLD Parser

ferrari2k:
## Topic and Scope

I added the possibility to parse more pages like picnic without parsing errors

## Concerns/issues

-/-

## Formal requirements

- [ x ] I did check that the app can still be opened and does not throw any browser logs
- [ x ] I created tests for newly added PHP code (check this if no PHP changes were made)
- [ x ] I updated the OpenAPI specs and added an entry to the API changelog (check if API was not modified)
- [ x ] I notified the matrix channel if I introduced an API change


christianlupus:
I am currently not at home and cannot work on this PR. I will give it a thorough review soon.

Generally speaking, enhancements are welcome. However, I want to be transparent here: the parser need a complete redesign. So, the most valuable part of this PR might be (well documented) tests with currently breaking sites/import dumps. That way, the reimplement procedure will keep this PRs functionality available in the long run. 
I just want to tell you this to avoid you be surprised/sad if I remove all old code from the parser soon. This should not discourage you nor should it stop you from working on the PR.

ferrari2k:
Hello @christianlupus, thanks for the reply, I'll be looking forward to an updated parser :)

PR #3124, opened 2026-05-02
Upgrade to Vue3

christianlupus:
## Topic and Scope

This is a technical update of teh Vue system to the new version 3. Some changes are needed as previously used Components were Vue2-only.

## Concerns/issues

It is just a WIP state at the moment.

## Formal requirements


christianlupus:
@seyfeb I started to work on the Vue 3 migration. Currently, this is not yet finished but I need to the CI checks now. I see that I need to do some reading to get the DCO happy. But the jest and some code styling I have to check individually. 

Just wanted to keep you posted. If you want to have a look prior to merging, just give me a ping. Otherwise, I would try to get this merged soon as any updated NPM packages cause collisions on package-lock.json. Thus, I would like to close this and eventually fix further issues down the road.

PR #3312, opened 2026-08-13
fix: shorten link symbol to other recipes

xhon-pelushi:
## Topic and Scope

Shorten the link-to-other-recipes symbol so the UI is less noisy when recipes reference each other.

Fixes #3068

## Concerns/issues

None expected — presentation-only change to the link symbol.

## Formal requirements

### DIFF
--- a/.changelog/current/3312-shorten-recipe-link-symbol.md
+++ b/.changelog/current/3312-shorten-recipe-link-symbol.md
@@ -1,0 +1,3 @@
+# Changed
+
+- Shorten the link symbol used when recipes reference other recipes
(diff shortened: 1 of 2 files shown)


christianlupus:
Please agree to the PR draft as the checkmark at the bottom should help with classifying the possible impact.

xhon-pelushi:
Thanks @christianlupus — filled in the PR template checklist and added `.changelog/current/3312-shorten-recipe-link-symbol.md`.

christianlupus (approved):

christianlupus:
Thanks for adjusting the PR.

There are failing tests and I do not understand just by looking at the overview what is causing this. I need to investigate a bit but I am on trip this week. So, either you directly see the problem or you will have to wait with this for next week. Then, I will get this sorted out, I guess.

xhon-pelushi:
Pushed a small follow-up to fix the Prettier CI failure on `src/js/title-rename.js` (the PHP linter job runs Prettier as its last step).

The separate “Block fixup and squash commits” failure looked like a GitHub Actions flake (could not download `block-autosquash-commits-action`); re-ran failed checks.

christianlupus:
Thanks for fixing the CI issues. Just merged the PR to master.

PR #3324, opened 2026-08-24
Mark as compatible with NC35

christianlupus:
## Topic and Scope

This PR marks the app as compatible with the upcoming NC 35.

## Concerns/issues

It is not thoroughly tested yet but unit tests should check most cases

## Formal requirements


christianlupus:
Current tests will fail as the `stable35` branch does not yet exist.

